PR #28, opened 2017-12-20
fix import from pydicom.tagtools to pydicom.tag

BrianKolowitz:


BrianKolowitz:
fix

vsoch:
hey @BrianKolowitz  - this PR looks good - I think for the continuous integration we need to update the version of pydicom used here --> https://github.com/pydicom/deid/blob/master/.travis.yml#L14 

I had specified it to that version explicitly because of maintaining the old version, but with this fix we should be good :)

BrianKolowitz:
I modified the .travis.yml to include the latest hash which makes the build pass. 
```
- pip install git+git://github.com/pydicom/pydicom.git@4ddd9a70000ec72a53ea2dab8bc8063043691548
```

vsoch:
Awesome! Huge thanks for doing this. I'm going to be opening up the development branch proper with a CHANGELOG, but these tweaks can go right into master for now. Merging!

vsoch:
Do you know why it shows the commits that I made from the (now merged) PR? I want to make sure I didn't muck something up with git - if those are in master the comparison shouldn't show them here.

BrianKolowitz:
I have to admit I'm not an expert in open source collaboration. I believe it kept tracking my commits along the way. I probably should have created a separate branch and opened the pull request from there. 

If you'd like to reject these changes I can create a clean pull request.

vsoch:
Ah, not necessary! Let's see what happens when I just merge, it should arguably show the same commits.

vsoch:
The code looks good! A few months back I think Github changed to a container based framework to run github pages and the site generally, so I'd bet they tweaked some algorithm that would update the code, etc. to re-test or re-compare against changed bases. In this case it's probably a bug, but I haven't seen it before so it's likely a one time thing. Anyway, thank you again! If you have other ideas for deid, please open an issue and we can chat about it. There aren't a ton of options out there for this task, and especially with python. I am hoping to (eventually) write a small paper on it, and you are welcome to be a part of that too!

PR #34, opened 2017-12-22
additional fixes from pull request #30

BrianKolowitz:
Hey I think we didn't propagate the fix to all of the files.


vsoch:
Ah thanks for fixing this! Yes I completely forgot about these other spots.

Issue #36, opened 2017-12-23
Break deid specification into its own class 


vsoch:
see discussion in #35

PR #37, opened 2017-12-24
Adding DeidRecipe configure class

vsoch:
This PR is based on discussion with @BrianKolowitz to add a proper DeidRecipe class that can handle loading (and then other operations) with deid recipe files. This first implementation has basic functions to load single, multiple, and control defaults, as well as retrieve / filter different actions and filters. The class is [shown here](https://github.com/pydicom/deid/blob/add/deid-recipe-class/deid/config/__init__.py) and a walk through of using it [provided as a script](https://github.com/pydicom/deid/blob/add/deid-recipe-class/examples/dicom/deid-dicom-example.py). I've also added updated docs (tutorial) for the cleaner (previously only existed as a script in the examples folder).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and changes prior to that are (unfortunately) done retrospectively. Critical ite
 
 ## [vxx](https://github.com/pydicom/deid/tree/development) (development)
 
+**additions**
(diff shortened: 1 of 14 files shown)


vsoch:
This PR will close https://github.com/pydicom/deid/issues/36

vsoch:
ping @BrianKolowitz

PR #38, opened 2018-01-04
Feature get files generator

BrianKolowitz:
I added 2 generator methods because deidentifying large datasets required loading the entire directory and dicom attributes into memory. You can see my revised example here https://github.com/BrianKolowitz/deid/blob/development/my_examples/dicom/my_deid.py


vsoch:
hey @BrianKolowitz ! I think this is a great idea - most definitely a generator is the right data structure for this use case. A suggestion - instead of having two (very similar functions), why not modify the current `get_files` and `recusive_find` to by default return an generator? For example:

```
files =get_files_iterator('/home/vanessa/Downloads',pattern='*')

$ files
$ <generator object get_files_iterator at 0x7f0bd4035468>
```

And I could then loop through it to reveal the files (using as generator):

```
for file in files:
    print(file)
```
but if I wanted a list, I would just as for it:

```
files = list(files)
```

Importantly, what we would want to do I think is to add the same validation function to the generator. This would mean validating a file before producing it for the user, and not producing it if not valid. For the list() part, it would mean (again) doing the check before returning the full thing, and you can use your preference for implementation.

Let me know your thoughts! The goal here would be to make the current functions more flexible to use cases over adding redundant code. And then of course we would want to make sure that we update the examples / general code base so that the user is made away of his/her options for using the function.

BrianKolowitz:
@vsoch I made the recommended changes.

vsoch:
Great! A few quick questions before ready for review:

 1. did you mean to add the .cache folder? What is that?
 2. the todo item: #` todo : consider using generator functionality`  I think given that this PR is adding the functionality, if this is related and makes the functions where you commented better integration with that new feature, we should add to the PR.

Super looking forward to testing this out!

vsoch:
ah and I'm just looking at your example - I will use it to also test the changes (and then make updates) and we can add it to examples for other users! I don't have a non linux (ubuntu) machine so the different OS is very useful.

vsoch:
hey @BrianKolowitz this looks good! I'm going to merge. Thanks for the contribution!

PR #41, opened 2018-01-08
Adding DeidRecipe Class (to development branch)

vsoch:
This PR is based on discussion with @BrianKolowitz to add a proper DeidRecipe class that can handle loading (and then other operations) with deid recipe files. This first implementation has basic functions to load single, multiple, and control defaults, as well as retrieve / filter different actions and filters. The class is shown here and a walk through of using it provided as a script. I've also added updated docs (tutorial) for the cleaner (previously only existed as a script in the examples folder).


vsoch:
hey @BrianKolowitz I think you wanted this feature / class, would you mind giving this PR a review?

PR #42, opened 2018-01-17
Fix save dicom

johannesu:


vsoch:
Oh wow good catch! Thank you!

Issue #44, opened 2018-01-19
Support for duplicate DICOM file names


johannesu:
In the function `deid.dicom.get_identifiers` the dicom files are identified only by the file name.

This leads to hard to find bugs if you are trying to deidentify several series at the same time.

E.g.
Suppose you have two series

````
dicom/seriesA/00001.DCM
dicom/seriesB/00001.DCM
````

And you wish to update the Series Instance UID. 

Following the example code you may write something like this 

```
ids = get_identifiers(dicom_files)
for image,fields in ids.items():    
    fields['instance_id'] = pydicom.uid.generate_uid(entropy_srcs=uid)
    updated_ids[image] = fields 
```

Then both images would get the same instance_id.

vsoch:
The full path to the file name would distinguish between seriesA and seriesB, the idea isn't to have those two images be the same, but different.

johannesu:
Maybe I misunderstand how the functions are meant to be used.  But `get_identifiers` remove the full path.
 
`item_id = os.path.basename(dicom_file)`

is used to index the files.

vsoch:
hey @johannesu ! Apologies for the delay, it's been a busy month. I agree about not removing the full path, and think that removing that line would resolve the issue. Would that resolve it for you as well? I can make a PR for you to test, if you like.

vsoch:
@johannesu just to be clear - you are suggesting to have get_identifiers use the full path as the index?

johannesu:
Yes, I think that would be a better solution.

johannesu:
Resolved in #46

PR #46, opened 2018-01-31
tweaking item ids to use full file name / path

vsoch:
@johannesu see this PR - this would change the `item_id` (the index for the identifiers) to not be the basename, but maintain the overall path. Would you like to test to see if it addresses #44 ?


johannesu:
Thanks!

If this does not produce any side effects (I am not sure how the item_id is used throughout the code). This would solve #44 .

Issue #47, opened 2018-03-21
Issue with full path names?


BrianKolowitz:
I updated to the latest version and my code broke. I'm using the full file path as the key for the ids dictionary. 

``` if idx in ids:``` breaks b/c idx is the basename but my keys are the full path

a) is this a bug?
b) (or) do i need to modify my code to remove the path from the ids keys

# header.py
```
if recipe.deid is not None:
            if idx in ids:
                for action in deid.get_actions():
                    dicom = perform_action(dicom=dicom,
                                           item=ids[idx],
                                           action=action) 
```

vsoch:
yes I think that is a bug - I just opened a PR with a fix to test! https://github.com/pydicom/deid/pull/48

PR #48, opened 2018-03-21
fixing bug in header.py with indexing

vsoch:
This will address #47 that introduced a bug to use the file name as index instead of the full path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,7 @@
 
 This is a manually generated log to track changes to the repository for each release. 
 Each section should include general headers such as **Implemented enhancements** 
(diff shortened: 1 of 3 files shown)


BrianKolowitz:
@vsoch looks like that fix worked but there's another bug. on windows you're replacing \\ with / when you reconstruct the file name in save_dicom. I pass in dir\\dir\\dicom.dcm and it returns dir\\dir/dicom.dcm

like 87 utils.py 

```
output_dicom = "%s/%s" %(output_folder,dicom_name)
```

vsoch:
Could you suggest a fix? I am guessing os.path.join or similar?

vsoch:
okay just pushed a fix, please test.

Issue #49, opened 2018-04-20
Example used with current source code gives ImportError


petkaze:
I checked out the current source tree, built the Docker image and then tried to use the data in the basic example. I get the exception detailed below when trying this, and it seems indeed like the import is missing in the code. 

```
➜  code git clone git@github.com:pydicom/deid.git
Cloning into 'deid'...
➜  ~ cd deid 
➜  deid git:(master) docker build -t pydicom/deid .
➜  deid git:(master) docker run pydicom/deid inspect --deid /home/peter/code/deid/examples/deid/deid.dicom /home/peter/code/deid/deid/data/dicom-cookies --save 

Traceback (most recent call last):
  File "/opt/conda/bin/deid", line 11, in <module>
    load_entry_point('deid==0.1.11', 'console_scripts', 'deid')()
  File "/opt/conda/lib/python3.6/site-packages/deid-0.1.11-py3.6.egg/deid/main/__init__.py", line 157, in main
    from .inspect import main
  File "/opt/conda/lib/python3.6/site-packages/deid-0.1.11-py3.6.egg/deid/main/inspect.py", line 30, in <module>
    from deid.dicom import get_files
  File "/opt/conda/lib/python3.6/site-packages/deid-0.1.11-py3.6.egg/deid/dicom/__init__.py", line 1, in <module>
    from .header import (
  File "/opt/conda/lib/python3.6/site-packages/deid-0.1.11-py3.6.egg/deid/dicom/header.py", line 29, in <module>
    from .tags import (
  File "/opt/conda/lib/python3.6/site-packages/deid-0.1.11-py3.6.egg/deid/dicom/tags.py", line 28, in <module>
    from pydicom.tag import tag_in_exception
ImportError: cannot import name 'tag_in_exception'
```

Any ideas what might be wrong?

vsoch:
Yes I think the pydicom version might be the newer one that doesn't have that tag. Could you check the version of pydicom installed?

petkaze:
This was executed from within the Docker container that is created when building from source. According to the Dockerfile that is part of the source tree, the following version of pydicom is installed 

RUN pip install git+git://github.com/pydicom/pydicom.git@affb1cf10c6be2aca311c29ddddc622f8bd1f810

vsoch:
okay definitely must be a bug then! Let me rebuild the image from scratch too so I can reproduce the issue and fix it up.

vsoch:
[...]

PR #50, opened 2018-04-23
fixing bug with tag saving

vsoch:
This will fix an  import bug with #49


vsoch:
this seems to be fixed up. The import was correct - the install in the container needed a version bump as suspected.

Issue #54, opened 2018-05-17
deid identifiers --action all fails with NoneType object has no attribute get_actions


petkaze:
I have pulled the current master branch and tried the following:
```
(deid) ➜  deid git:(master) ✗ cd examples/dicom 
(deid) ➜  dicom git:(master) ✗ python deid-dicom-example.py 
Found 1 valid dicom files
Found 1 valid dicom files
Found 1 valid dicom files
Found 1 valid dicom files
Found 1 valid dicom files
Found 1 valid dicom files
Found 1 valid dicom files
WARNING No specification, loading default base deid.dicom
WARNING No specification, loading default base deid.dicom
WARNING No specification, loading default base deid.dicom
Traceback (most recent call last):
  File "deid-dicom-example.py", line 214, in <module>
    output_folder='/home/vanessa/Desktop')
  File "/home/peter/code/deid/deid/dicom/header.py", line 272, in replace_identifiers
    for action in deid.get_actions():
AttributeError: 'NoneType' object has no attribute 'get_actions'
```

Some information about my environment:
```
(deid) ➜  dicom git:(master) ✗ python --version
Python 3.6.2
(deid) ➜  dicom git:(master) ✗ pip list
certifi (2018.4.16)
chardet (3.0.4)
cycler (0.10.0)
deid (0.1.13, /home/peter/code/deid)
idna (2.6)
kiwisolver (1.0.1)
matplotlib (2.2.2)
numpy (1.14.3)
pip (9.0.1)
pydicom (1.0.2)
Pygments (2.2.0)
pyparsing (2.2.0)
python-dateutil (2.7.3)
pytz (2018.4)
requests (2.18.4)
retrying (1.3.3)
setuptools (28.8.0)
simplejson (3.15.0)
six (1.11.0)
urllib3 (1.22)
validator.py (1.2.5)
```

Any ideas what might be wrong?

vsoch:
Any reason your output folder is still set to my Desktop?

petkaze:
@vsoch Probably because the deid-dicom-example.py file has your Desktop as the default value. See for example lines 214 and 219.

vsoch:
The general idea is that you should modify the example for your needs. Please modify the paths and try again, and run the script relative to its location as I suggested above. Thanks!

vsoch:
Awesome! Just merged :) Let me know if you have any more issues, I'll close the issue if you are resolved (reopen if needed!)

PR #55, opened 2018-05-21
Correct the supplied example and make it more generic.

petkaze:
Fixes for https://github.com/pydicom/deid/issues/54

Always supply the recipe to replace_identifiers.
Use a more generic path as default argument.


vsoch:
Thanks! I can't believe I hard coded a path to my Desktop in there, this is definitely a #vanessafail :)

PR #56, opened 2018-06-05
Support passing Dataset objects instead of requiring FileDataset objects

petkaze:
We typically operate on pydicom.Dataset objects, and everything is not necessarily files on disk. This PR removes the requirement to provide a filename for all Dataset objects, thay may be streams that exist in memory and are not directly referenced as a file on disk.

Since the filename variable is only used for logging purposes, this can be removed without losing any features.

### DIFF
--- a/deid/dicom/actions.py
+++ b/deid/dicom/actions.py
@@ -86,7 +86,6 @@ def _perform_action(dicom,field,action,value=None,item=None):
     Both result in a call to this function. If an action fails or is not
     done, None is returned, and the calling function should handle this.
     '''
-    dicom_file = os.path.basename(dicom.filename)    
     if action not in valid_actions:
         bot.warning('%s in not a valid choice [%s]. Defaulting to blanked.' %(action,
                                                                               ".".join(valid_actions)))
--- a/deid/dicom/fields.py
+++ b/deid/dicom/fields.py
@@ -83,7 +83,6 @@ def get_fields(dicom, skip=None, expand_sequences=True):
         skip = [skip]
     fields = dict()
     contenders = dicom.dir()
-    dicom_file = os.path.basename(dicom.filename)
     for contender in contenders:
         if contender in skip:
             continue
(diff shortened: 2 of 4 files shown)


deid/dicom/actions.py
```diff
@@ -86,7 +86,6 @@ def _perform_action(dicom,field,action,value=None,item=None):
     Both result in a call to this function. If an action fails or is not
     done, None is returned, and the calling function should handle this.
     '''
-    dicom_file = os.path.basename(dicom.filename)    
     if action not in valid_actions:
```
vsoch: This looks good - this variable isn't used anywhere anymore.

deid/dicom/fields.py
```diff
@@ -83,7 +83,6 @@ def get_fields(dicom, skip=None, expand_sequences=True):
         skip = [skip]
     fields = dict()
     contenders = dicom.dir()
-    dicom_file = os.path.basename(dicom.filename)
     for contender in contenders:
```
vsoch: Ditto.

petkaze:
Sweet, thanks for the rapid feedback! :+1: 

I have run some basic tests locally with this and thing seems fine. However, it would be nice to look into creating a unit test suite that can be integrated into the Travis build at some point.

vsoch:
Agreed! I'm not sure why the testing didin't trigger for this PR - the webhook seems to work with master. I'm going to merge and see if that triggers it, and we can add tests at some point if necessary / someone has bandwidth.

vsoch:
Okay, there was one bug with the format string but I fixed it up. https://travis-ci.org/pydicom/deid/builds/388335580 I'm not sure why the CI wasn't triggering but if it happens again I'll try reconnecting. It seems to still be set as a service (and not webhook) so maybe it's deprecated and the connection needs to be made again.

petkaze:
Nice, thanks a lot @vsoch

PR #61, opened 2018-09-05
Fixed region anonymization (vertical vs horizontal coordinates)

fimafurman:
Before the bug is fixed, notice that the area on the left is blacked out, however, PHI information is across the top of the screen (I did blur it, as this is a real image)
badly_cleaned

After this proposed fix, the black box correctly blanks out the top region of the image leaving the rest of the image intact.
cleaned_correctly

### DIFF
--- a/deid/dicom/pixels/clean.py
+++ b/deid/dicom/pixels/clean.py
@@ -119,7 +119,7 @@ class DicomCleaner():
 
             for coordinate in coordinates:
                 minr, minc, maxr, maxc = coordinate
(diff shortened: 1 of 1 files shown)


vsoch:
Is it ok with you if I fork your PR to fix the CI issues? We need to set the version of Matplotlib. Your commits will be preserved!

fimafurman:
Yes, of course! When do you think the fix could get into the master?

vsoch:
haha, let me see how quickly I can figure out this bug :_) Hang tight!

vsoch:
Closing, commits have been integrated via #62

PR #80, opened 2018-12-10
Update/deid.dicom

### LINKED ISSUE #63: Is deid.dicom aquivalent to CTP's pixel.anonymizer.script?
I noticed that deid.dicom is missing entries in CTP's pixel.anonymizer.script - is deid's deid.dicom config file modeled after CTP's script?

vsoch:
# Description

This will close #63, and add a list of contributions from @fimafurman.  The contributed file is here:
https://github.com/pydicom/deid/issues/63#issuecomment-445898849 and those newly added have comment (FFUR) or (AMBR). The main difference from the original ordering is that I placed those in the blacklist last, allowing images to pass through first that would be flagged for other reasons and possibly able to be anonymized. This of course will vary based on the use case, etc.

@fimafurman is there any reason you have several listings of, for example, blacklist in your file? They are read in as an ordered dict, so having an entry for the blacklist first would place it at the top (even for those that come later).

Could you take a look at let me know if the PR looks okay?

And what are your thoughts generally on this task? I personally think character detection will go a lot farther, and likely all the big companies are working on it.  But in the meantime given that the user has a reasonable list for his/her possible, this is probably better than nothing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - updating deid.dicom with contribution from @fimafurman [#63](https://github.com/pydicom/deid/issues/63) (0.1.22)
  - adding "func" option for recipe to pass function (0.1.21)
  - fixing client bug, redoing docs to be better organized (0.1.20)
  - Removing MediaStorageSOPInstanceUID from file_meta, issue #72 (0.1.19)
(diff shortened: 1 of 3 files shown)


fimafurman:
Perhaps, it might be best to remove my blacklist entries as those may be specific to our case of never wanting any images in certain categories. 

I suspect that character detection will indeed be the way companies solve this in the future. Google is working on Health API and Amazon, I am sure, will catch up as well. OCR/NLP based detection may not always work 100% and it is slower, so I could envision both methods working in combination. For example, images that do not trigger a rule in deid, maybe routed for OCR/NLP based detection which would result in a faster and more accurate processing than either of the methods in isolation.

PR #85, opened 2018-12-27
jitter_timestamp with variable name change

howardpchen:
JITTER command in recipe generates an error that points to this file.
jitter_timestamp was called in line 121 with "item=dicom" but the
function itself now asks for parameter name "dicom" so changed line 121
to dicom=dicom

Also the code for jitter adding datetime together with simple + in line
160 feels unusual for date operation.  Added new code to utilize
datetime package to do the jitter and updates the dicom field.

In dicom/utils/action.py there is a get_timestamp func which is imported
by actions.py and seems to deal with jitter but looks to be heavier than
necessary since JITTER operates in +/- days only based on Documentation.

Neither current nor old jitter_timestamp uses the get_timestamp
function.

# Description

Related issues: #86 

# Checklist

# Open questions

### DIFF
--- a/deid/dicom/actions.py
+++ b/deid/dicom/actions.py
@@ -183,7 +183,7 @@ def jitter_timestamp(dicom, field, value):
             new_value = None
             bot.warning("JITTER not supported for %s with VR=%s" % (field, 
                                                                     dcmvr))
-        if (new_value is not None and new_value is not original):
+        if (new_value is not None and new_value != original):
             # Only update if there's something to update AND there's been change
             dicom = update_tag(dicom,
                                field=field,
--- a/deid/dicom/actions.py
+++ b/deid/dicom/actions.py
@@ -166,7 +166,7 @@ def jitter_timestamp(dicom, field, value):
         '''
         if (dcmvr == 'DA'):
             '''
-             NEMA-compliant format for DICOM date is YYYYMMDD
+            NEMA-compliant format for DICOM date is YYYYMMDD
             '''
             new_value = get_timestamp(original, jitter_days=value, 
                                       format='%Y%m%d')
@@ -180,10 +180,12 @@ def jitter_timestamp(dicom, field, value):
                                       format='%Y%m%d%H%M%S.%f%z')
         else:
             # Do nothing and issue a warning.
-            new_value = original
+            new_value = None
             bot.warning("JITTER not supported for %s with VR=%s" % (field, 
                                                                     dcmvr))
-        dicom = update_tag(dicom,
-                           field=field,
-                           value=new_value)
+        if (new_value is not None and new_value is not original):
+            # Only update if there's something to update AND there's been change
+            dicom = update_tag(dicom,
+                               field=field,
+                               value=new_value)
     return dicom
--- a/deid/dicom/actions.py
+++ b/deid/dicom/actions.py
@@ -157,12 +157,33 @@ def jitter_timestamp(dicom, field, value):
         value = int(value)
 
     original = dicom.get(field,None)
-    original_date = datetime.datetime.strptime(original, "%Y%m%d")
-    date_added = original_date + datetime.timedelta(days=value)
-       
+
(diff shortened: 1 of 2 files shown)


deid/dicom/actions.py
```diff
@@ -156,6 +157,12 @@ def jitter_timestamp(dicom, field, value):
         value = int(value)
 
     original = dicom.get(field,None)
+    original_date = datetime.datetime.strptime(original, "%Y%m%d")
```
vsoch: This is going to trigger an error if the field is not in the dicom, or the format doesn't match the one you have provided. What is the rationale for making it more complicated than adding the number of days to the original?

vsoch:
hey @howardpchen this PR is really great and I'm so glad you contributed it! I'll have some time tomorrow to give it a proper review, so stay tuned! :L)

vsoch:
I think this was a casualty of when I tried to reorganize the functions, removing the identifiers module that had the right "get_timestamp" function, here --> https://github.com/pydicom/deid/blob/38c6ed91591ba5a340079d6ec4b3c52107e03348/deid/identifiers/utils.py

Take a look at the original function, I'll put it here too:

```python
def get_timestamp(item_date, item_time=None, jitter_days=None, format=None):
    '''get_timestamp will return (default) a UTC timestamp 
    with some date and (optionall) time. A different format can be 
    provided to change default behavior. eg: "%Y%m%d"
    '''
    if format is None:
        format = "%Y-%m-%dT%H:%M:%SZ"

    if item_date in ['',None]:
        bot.warning("No date in header, cannot create timestamp.")
        return None

    if item_time is None:
        item_time = ""

    timestamp = dateutil.parser.parse("%s%s" %(item_date,item_time))
    if jitter_days is not None:
        jitter_days = int(float(jitter_days))
        timestamp = timestamp + timedelta(days=jitter_days)

    return timestamp.strftime(format)
```
Actually, it's still there, under deid.utils.actions. The issue is with the jitter_timestamp function, which also used to be the following:

```python
# Timestamps
def jitter_timestamp(field,value,item):
    '''if present, jitter a timestamp in dicom
    field "field" by number of days specified by "value"
    The value can be positive or negative.
    '''
    value = to_int(value)
    original = item.get(field,None)
    if original is not None:
        jittered = get_timestamp(item_date=original,
                                 jitter_days=value,
                                 format="%Y%m%d")
        bot.debug("JITTER %s + (%s): %s" %(original,
                                           value,
                                           jittered))
        item[field] = jittered
    return item
```
This would explain why "item" is the argument instead! Would you care to test this (original) function and see if it fixes the issue?

vsoch:
I can also quickly do the change and open a PR for you to test, whatever is easier for you!

vsoch:
Actually your solution is essential the same strategy (but better because you check for the format!)

deid/dicom/actions.py
```diff
@@ -156,6 +157,33 @@ def jitter_timestamp(dicom, field, value):
...
+            YYYYMMDDHHMMSS.FFFFFF&ZZXX
+            '''
+            new_value = get_timestamp(original, jitter_days=value,
+                                      format='%Y%m%d%H%M%S.%f%z')
+        else:
+            # Do nothing and issue a warning.
+            new_value = original
+            bot.warning("JITTER not supported for %s with VR=%s" % (field, 
+                                                                    dcmvr))
+        dicom = update_tag(dicom,
```
vsoch: This looks great! My last comment is here - if we haven't changed the value, then we don't need to update the tag. I would set new_value to None, and then only update_tag if new_value isn't equal to None. You could also check if the new_value is equal to the original (and not update the tag if so). Does that make sense? Beautiful commenting here, by the way!

vsoch:
Great work! Merging.

vsoch:
https://pypi.org/project/deid/0.1.25/

And @howardpchen I added you as a contributor! https://github.com/pydicom/deid/blob/master/AUTHORS.md

PR #88, opened 2018-12-30
first go to add extra field expanders to close #87

### LINKED ISSUE #87: Additional field expanders
My group's use case includes passing a function to all DICOM Tags to check for certain PHI words/phrases.  Noticed that while we can do startswith and endswith, there isn't an easy way to apply func:my_function on most or all of the DICOM tags.  Would it be of interest if I expanded the deid/dicom/fields.py functionality (e.g. add expander words like "all," "contains," "allexcept" etc) and sent out a pull request?  Thanks!

vsoch:
# Description

This will resolve https://github.com/pydicom/deid/issues/87, adding extra expanders to the selection of fields, including:

 - allfields: to select all fields
 - exceptfields: to select a pipe (|) separated list of fields to not include
 - contains: to select fields based on containing a string (or multiple with pipe)

I have also updated the docs to show examples of the above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
- - added all, allexcept, contains to field filters (0.1.24)
+ - added all, allexcept, contains to field filters [#87](https://github.com/pydicom/deid/pull/87) (0.1.26)
+ - fixing bug in jitter timestamp function [#85](https://github.com/pydicom/deid/pull/85) (0.1.25)
  - installation order of pydicom / matplotlib changes default python [#81](https://www.github.com/pydicom/deid/issues/81) (0.1.23)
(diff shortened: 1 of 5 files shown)


vsoch:
One question I have is if we should do `allfields` and `exceptfields` vs `all` and `except`. I did the first per your suggestion, but I think the second reads better. if we need to add another distinguisher to all or except we could still do that (e.g., all<something>) and I can't think of a good example off the top of my head. Maybe allprivate or similar!

howardpchen:
Hey @vsoch agree `exceptfields` reads better.

Something like `allprivate` might be useful if there are some known PHI issue worth running a sophisticated PHI check only on the private tags.  For example, for our use case, we're searching for elements of Patient Name tag within all the Data Elements we're keeping because one of our vendors uses a private tag for patient name.  There are also private tags where it looks like the scanner baked in human input because the patient name or address is misspelled... those are a lot trickier but almost exclusively reside in private tags.

For now I'm just using the remove_private_tags in pydicom and `replace_identifiers(... remove_private=True)` to go nuclear on all the private tags.

Back to the topic on hand. The code looks good, and thanks for also updating the documentation and version log.

vsoch:
okay, I'm good with `allfields` and `exceptfields`, I think I'll add `all` to allow the user to specify `ALL` as an option too.

vsoch:
https://pypi.org/project/deid/0.1.26/

PR #99, opened 2019-01-18
fixing bug with not contains, is also expecting a value

### LINKED ISSUE #98: notContains not expecting a field?
I had to use 'notContains' clause in deid rules and expected it to work just like contains clause with the opposite effect, however, I found that the parser does not actually expect the value to be present. It looks like a bug, as notcontains should expect a field value:

In utils.py:

`# Missing, empty, notcontains expect only a field
 elif action in ['missing', 'empty','notcontains', 'present']:
       field = member.strip()`

vsoch:
This will close #98, specifically the "notcontains" action should behave similar to contains, but with the opposite effect. Specifically,

```
notcontains <field> <expression>
```

### DIFF
--- a/deid/config/utils.py
+++ b/deid/config/utils.py
@@ -347,16 +347,16 @@ def parse_member(members, operator=None):
         action, member = member.split(' ',1)
         action = action.lower().strip()
 
-        # Contains, equals, not equals expects FieldName Values
-        if action in ['contains','equals','notequals']:
+        # Contains, notcontains, equals, not equals expects FieldName Values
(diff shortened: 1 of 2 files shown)


vsoch:
@fimafurman if you'd care to test and verify, I'll merge and release when you've confirmed.

vsoch:
https://pypi.org/project/deid/0.1.32/

fimafurman:
Thank you for taking care of this so quickly!


-- 

Fima Furman

617-512-3462

Director of Machine Learning and AI

ffurman@ambrahealth.com



KLAS Category Leader for
Image Exchange since 2014

Issue #102, opened 2019-06-23
PHI field is not replaced when inside sequence


danielsnider:
With recipe:
```
REPLACE InstanceCreationDate func:generate_uid
```

The field `InstanceCreationDate` is not replaced when inside of a sequence. See how there are two occurences of InstanceCreationDate and only one got de-identified? My result as seen by `dcmdump`: 

![image](https://user-images.githubusercontent.com/1737560/59978667-c4317580-95ac-11e9-881e-95cb92f8e4ec.png)

What do you think?

Thanks!
Daniel Snider ツ

vsoch:
I think a check needs to be done for a sequence, and if it's a sequence, iterate over it. But that's just me :)

vsoch:
@danielsnider do you have a dummy dicom I can use to test?

danielsnider:
Yes. One minute!

danielsnider:
@vsoch [here](https://sickkidsca-my.sharepoint.com/:u:/g/personal/daniel_snider_sickkids_ca/EU6ypKt3WqdIjEXhWu0mSZQB4pTo-aw0tMCYImOK-9jdTQ?e=hABXMA) you go. Thanks so much.

danielsnider:
It is the DICOM from my example, look for `InstanceCreationDate`.

vsoch:
How do you propose we do this? Currently, it's a matter of looking in the dicom to see if a tag is present. You are asking to additionally parse every tag to check for sequences, and further, to do this recursively. This would hugely slow down the library to need to look everywhere. I don't think this would be feasible, because what if it's a sequence in a sequence? How deep do we go?

Instead, I'd propose that a user can define a nested tag, and that the choice is explicit. So in your case you would need to do:

```
REPLACE ReferencedPerformedProcedureStepSequence.InstanceCreationDate func:generate_uid
```

vsoch:
This library also badly needs pylint, it's driving me nuts just looking at it right now, haha. Based on how easy / hard this fix is, I might pylint first, and maybe create a conda-forge recipe too.

danielsnider:
[...]

PR #106, opened 2019-06-24
Add/sequence replacement

vsoch:
# Description

Related issues: #102 

This will allow for REPLACE and BLANK to work for nested sequences.  I've included a full example in the docs, and linked to an example folder in the repository.


vsoch:
Here you go @danielsnider, there is a full example added in the docs and examples folder. This will cover `REPLACE` and `BLANK` since we can easily dig into the nested structure and edit it (and thanks to Python objects the parent also gets the changes) but I didn't see an easy way to remove an item from a sequence, or ADD. If someone else wants these in the future, they can complain about it and PR. :)

danielsnider:
I'm going to test it right meow! <3

danielsnider:
Works like a charm! I'm reviewing the code some more too. 👍

Thanks so much. If you're ever in Toronto I'll see to it that your treated like royalty.

vsoch:
Haha, nice!

While you’re reviewing, if you have ideas for how we could also support commands to add or remove content, this could be desired say day. I tried the traditional del varname to see if it might, for example, remove a DataElement from a sequence, but it didn’t work. The only way would be to loop through the sequence, and then not add an element to some new sequence. But then we would have to keep the parent as well to link to it.

Is it worth trying, or save for some future user to ask about?

danielsnider:
I would say fix it when truly needed 👍

vsoch:
Sounds good. Let me know when you've finished your review and I'll release.

danielsnider:
Will do!

danielsnider:
Approved 👍. I learned a lot from reading your code. I don't have any more suggestions, sorry!

vsoch:
No worries! I just pushed a fix for the typo in the docs. When this passes, I'll merge and then release on pypi (and then after that, it gets PRd --> released on conda too!)

vsoch:
Always a pleasure @danielsnider !

https://pypi.org/project/deid/0.1.34/

danielsnider:
Huge thanks @vsoch !

PR #110, opened 2020-02-08
Adding black linter, testing github actions for pydicom

vsoch:

This PR will add black (and a check for it using GitHub actions) to ensure that the formatting is consistent. I'm also updating the docstrings.


vsoch:
okay - so tests are passing with previous versions, this does not mean it will work to install just with conda (likely dependency conflicts) so we will try another PR to test that.

Issue #109, opened 2020-02-07
Update version of Matplotlib 


davidparsson:
Is there a particular reason behind the specified max version of matplotlib requred?

I'm trying to install this in a project that's using a newer version of matplotlib, and a dependency management tool that does not allow version mismatches.

Thanks!

vsoch:
There was a particular reason - but I didn't document it and I can't remember. I think it had something to do with the cleaning of pixels functionality, and some function afforded by previous versions of matplotlib.

Would you care to remove the version dependency, and test out the library, and report back any issues?

davidparsson:
I don't really have a lot of time, nor deid experience, but if I find the
time I might give it a try.

Den fre 7 feb. 2020 kl 23:23 skrev Vanessasaurus <notifications@github.com>:

vsoch:
hey  @davidparsson I'll see if I can make this slightly easier for you with a branch that explicitly removes the version dependency. Give me a little bit and I'll push - also going to add black for formatting and a few (non functional change) updates.

vsoch:
okay just hit the error with local testing - a deprecated key "scale"
```
======================================================================
ERROR: test_add_tag (test_dicom_tags.TestDicomTags)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/vanessa/Desktop/Code/deid/deid/tests/test_dicom_tags.py", line 49, in test_add_tag
    from deid.dicom.tags import add_tag
  File "/home/vanessa/Desktop/Code/deid/deid/dicom/__init__.py", line 12, in <module>
    from .pixels import has_burned_pixels, DicomCleaner
  File "/home/vanessa/Desktop/Code/deid/deid/dicom/pixels/__init__.py", line 2, in <module>
    from .clean import DicomCleaner
  File "/home/vanessa/Desktop/Code/deid/deid/dicom/pixels/clean.py", line 34, in <module>
    matplotlib.use("pdf")
  File "/home/vanessa/anaconda3/lib/python3.7/site-packages/matplotlib-3.2.0rc3-py3.7-linux-x86_64.egg/matplotlib/cbook/deprecation.py", line 296, in wrapper
    return func(*args, **kwargs)
  File "/home/vanessa/anaconda3/lib/python3.7/site-packages/matplotlib-3.2.0rc3-py3.7-linux-x86_64.egg/matplotlib/cbook/deprecation.py", line 358, in wrapper
    return func(*args, **kwargs)
[...]

PR #111, opened 2020-02-08
remove restriction on matplotlib (results in error)

vsoch:
This pull request is opened to provide a base for testing / debugging the current issue with updating matplotlib. Specifically, the problem is issue #109 - by setting a max (working) version you have version conflicts when installing with conda. The branch here removes the dependency, and I welcome testers to take it for a spin before we decide to merge proper.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -32,6 +32,7 @@ jobs:
           conda create --quiet --name testing
           export PATH="/usr/share/miniconda/bin:$PATH"
           source activate testing
+          pip install matplotlib
           python setup.py install
 
       - name: Test deid
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
     python: "2.7"
 
 install:
-  - pip install pydicom
+  - pip install pydicom matplotlib
   - cd $TRAVIS_BUILD_DIR/
   - python setup.py sdist
   - python setup.py install
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - Removing matplotlib dependency (0.1.36)
  - Adding black formatting, tests run in GitHub actions (0.1.35)
  - Adding option to recursively replace sequences (0.1.34)
  - adding pylint to clean up code (0.1.33)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,9 @@
 FROM continuumio/miniconda3
 
-RUN apt-get update && apt-get install -y wget git pkg-config libfreetype6-dev
-RUN /opt/conda/bin/conda install matplotlib==2.1.2
-RUN pip install pydicom==1.1.0
-RUN mkdir /code
-ADD . /code
+RUN apt-get update && apt-get install -y wget git pkg-config libfreetype6-dev g++
+RUN conda install matplotlib
 WORKDIR /code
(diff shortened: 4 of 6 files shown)


vsoch:
This has been tested in #109 and I'm approving merge into a release.

davidparsson:
I'm noting that this PR does not completely remove the version restriction, but rather reverses it. Is that intended to be merged as is?

vsoch:
I'm confused - you reported that you tested this. What exactly did you test?

vsoch:
The pull requests sets a minimum version of the known supported version, which would allow for installing anything above that. This is a change from the previous == requirement.
```
INSTALL_REQUIRES = (
    ("matplotlib", {"min_version": "2.1.2"}),
    ("pydicom", {"exact_version": "1.2.1"}),
    ("python-dateutil", {"min_version": None}),
)
```
Could you please provide detail on 1) what you tested if it wasn't this, and 2) what solution you expected here.

davidparsson:
This PR is what I tested, and indeed it sets a minimum version. I think it made sense to set the minimum version to allow testing it properly, but what I expected to be released was a change that completely remove the version restriction.

I believe that libraries like this one should try to be as permissive as possible when it comes to versions of dependencies, to make life easier for projects relying on this one, and this change removes support for a lot of versions that was allowed before without any obvious reason.

That being said, I am personally all right with this being released since I want the newer version.

vsoch:
I respect that you have personal beliefs, but I need more than that to completely lift a minimum version requirement. A version less than 2.1.2 would be very old, and I don't have any evidence that this would work. It's already a little shaky lifting the exact requirement, and I am almost sure we will hit some point in the future when an update to matplotlib breaks something here (and I'll need to address that then). This is just the burden of maintaining software - it's a living thing that is constantly changing, and undoubtedly there will be another user in time that reports that lifting the requirement has broken his/her build.

vsoch:
I see your point though - it's completely reversed. I'm also not comfortable with this now that you've pointed it out - even if it would work for new installs, it might break older ones that already have an older matploblib. This is very messy, but I'll do another small point release that lifts it entirely.

Next time, could you please properly review the PR, or report that you did not? The miscommunication has led to this.

davidparsson:
Regarding the miscommunication. I really think you'd have to be very clear if you expect a code review from anyone that's not a maintainer as that's not common practice in my experience. But I'm sorry if I've caused any trouble.

Thanks again for being helpful, and for timely responses and releases!

vsoch:
I think I was [pretty clear](https://github.com/pydicom/deid/issues/109#issuecomment-586365685), I explicitly said "test the PR."

> hey @davidparsson any updates on being able to test the PR? I also wanted to get your feedback on using matplotlib-base see the conda suggestion here! Would you have time in the coming future to test either of these for your use case?

What do you need me to say next time?

vsoch:
And there is no golden rule that contributors, however they might contribute (a question / request via issue or PR) should not also be helpful. In this case (and actually many cases) I am the _only_ maintainer and I ask for help because I need it. I don't have a team of folks I can turn to (the "other maintainers") that have taken on the responsibility for testing. It's just me.

davidparsson:
> And there is no golden rule that contributors, however they might contribute (a question / request via issue or PR) should not also be helpful.

Ok, that's not fair. Is that your response to [constructive criticism](https://github.com/pydicom/deid/blob/master/CONTRIBUTING.md#our-standards)? I'm the sole maintainer of several projects so I'm aware of the burden, so I've done my best to be both helpful and polite. I've spent a few hours on verifying the functionality of the PR (according to my definition of "testing"), but if you don't think that's helpful then I'm sorry.

vsoch:
That was very helpful, thank you. Have a nice day!

davidparsson:
You too! I really appreciate the quick resolution and release!

PR #113, opened 2020-03-02
Add REMOVE with custom function to screen tags

vsoch:
This PR is start of work to address #112, specifically we need a way to provide a custom function to look over all tags, and remove based on some condition. As is described in the documentation, this will require providing a function to the item that will return True or False.

This PR does not yet include adding private tags to list of fields, as this needs more testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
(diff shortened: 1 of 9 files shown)


vsoch:
I can definitely change to bot.warning. As for the error on return types, this occurred to me when I was writing the functions for REMOVE that require a boolean return. I do think we should address this, but not scoped to this PR. I'll open an issue for it.

vsoch:
okay take a look - I changed 3/4 of the bot.exit to bot.warning. I kept bot.exit in the case that REMOVE is used incorrectly (something other than a func is provided, which REMOVE doesn't currently support).

vsoch:
Thank you for your review! https://pypi.org/project/deid/0.1.38

Issue #123, opened 2020-03-12
Add additional filters for REMOVE


vsoch:
Now that REMOVE has support for a boolean func, it makes sense to also add the other filters (as they return booleans)

```
REMOVE <field> contains:expression
REMOVE <field> notcontains:expression
REMOVE <field> equals:expression
REMOVE <field> missing
REMOVE <field> notequals:expression
REMOVE <field> present
REMOVE <field> empty
```

vsoch:
Closed with #124

Issue #119, opened 2020-03-06
Discussion: Add class to handle get/set


vsoch:
Right now, the user is required to run the get and replace identifiers functions separately. Between these actions, there is a step to add any variables or functions that are needed. My thinking is that we might be able to provide an easier to use interface with a class. It might look something like:

```python
from deid.dicom import HeaderParser
parser = HeaderParser("dicom.deid")

# Add a function to some lookup to be used for replace
# name would be optional (for the key) and default to the function's name
parser.add_func(func=replace_uid, name="replace_uid")

dicom = pydicom.read_file(dicom_file)

for dicom in dicom_files:
    parser.run(dicom)

# And run would correspond to running get() and set() for get and replace identifiers, respectively.

def run(self, dicom):
    self.get(dicom)
    self.replace(dicom)
```
of course with many more variables than I'm adding here. and careful thought about how to store the fields, functions, added variables, etc.

vsoch:
@wetzelj using [iterall](https://github.com/pydicom/pydicom/blob/e7de71b4a2c7e15b1d23197a00642951cd695d94/pydicom/dataset.py#L2010) might be useful instead of dicom_dir().

vsoch:
Here are some example of iterating through things. When we use iterall we get a DataElement each time, all of which have a name:

```python
(0008, 0100) Code Value                          SH: 'T-D3000'
(0008, 0100) Code Value                          SH: 'R-10206'

In [111]: for element in dicom.iterall(): 
     ...:     if  "Code Value" in element.name: 
     ...:         print(element) 
```
but not all of them have what we traditionally use, keywords (e.g. PatientID or PixelData). The above elements (unwrapped from sequences) do not.

```python
In [111]: for element in dicom.iterall(): 
     ...:     if not element.keyword: 
     ...:         print(element) 
     ...:                                                                                                                                                                
(0011, 0003) Private Creator                     AE: 'Agfa DR'
(0019, 0010) Private Creator                     LO: 'Agfa ADC NX'
(0019, 1007) Private tag data                    CS: 'YES'
(0019, 1021) Private tag data                    FL: 6.039999961853027
(0019, 1028) Private tag data                    CS: 'NO'
(0019, 1030) Private tag data                    LT: ''
(0019, 10f5) [Cassette Orientation]              CS: 'LANDSCAPE'
(0019, 10fa) Private tag data                    IS: "297"
(0019, 10fb) Private tag data                    FL: 2.4000000953674316
[...]

Issue #118, opened 2020-03-06
Apply deid recipe actions to private DICOM tags


wetzelj:
As discussed within PR #112 (comments), expand the list of fields evaluated to include private tags.  This should enable recipe actions to be defined and applied to private tags in the same manner as they can be applied to public tags. 

Since private tags do not have an assigned Name, these will have to be identified using the DICOM tag numbers.

vsoch:
Yes! And we'll also need to better account for types. Most private tags can be indexed with their tag object, but when it comes to assigning some new value, we can't just throw in a string like we would for something like PatientID. This is similar to the function assignment issue, but more common for tags because it's never a simple string, but usually another Data Element.

vsoch:
This is definitely done! Please open a new issue if anything comes up.

PR #124, opened 2020-03-12
Adding ability to specify a filter with REMOVE

vsoch:
Currently, REMOVE <field> can handle a func:, however since we are returning a boolean, we can easily also support the already existing filters that return a boolean (contains, equals, etc.) This
change will allow for those filters. This will address #123, pinging @wetzelj 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - adding filters (contains through missing) for REMOVE (0.1.41)
  - adding support for tag groups (values, fields) (0.1.4)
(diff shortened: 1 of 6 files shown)


vsoch:
Yeah I was just thinking about this too - I was grabbing the filters from the list already derived, but I don't think those particular filters make sense. I'm going to remove them.

vsoch:
oops didn't mean to close it.

vsoch:
https://pypi.org/project/deid/0.1.41/

vsoch:
@wetzelj let me think about your comment:

> Regarding the plan forward... #118 is really one of the higher priority issues from our standpoint, but if you think that #119 should be done first, that's okay too.

In that adding private tags will mean parsing over the tag objects (which reveal data elements) I'm actually thinking that these two separate issues need to be addressed at the same time - we will be able to easily loop over private tags if the get/set is available. I can't promise starting on this today, but soon!

wetzelj:
No problem!  I appreciate all of your work on this!

Issue #128, opened 2020-05-14
Issue with a DICOM file that contains a list in the ImageType element


SpiroGanas:
Hello,

One of my DICOM files has an ImageType element that is a list:

![image](https://user-images.githubusercontent.com/3968834/81886707-5f3b9100-956b-11ea-8d15-83d7890320d0.png)

Could you please update line 158 of filter.py to handle this situation?

https://github.com/pydicom/deid/blob/8ca068ea17a6c7c2997c12105c1b9ae1f460470b/deid/dicom/filter.py#L152-L163

Thank you very much,

vsoch:
Happy to! Could you please provide a file and code snippet to reproduce the bug, and the version of deid? I’ll put in some time tomorrow. Thank you!

SpiroGanas:
deid version: 0.2.0

This code adds a few fields to your example DICOM file, and then runs the code that gave me an exception:

``` Python
import pydicom
from deid.dicom import get_files 
from deid.data import get_dataset
from deid.dicom import DicomCleaner

base = get_dataset('dicom-cookies')
dicom_files = list(get_files(base)) 
dicom_file = dicom_files[3]

ds = pydicom.dcmread(dicom_file)

 
# An exception occurs in the deid test file because there is no manufacturer element in the dicom
# So I added this to silence that error
ds.Manufacturer = 'TOSHIBA'

 
# Add the multi-value ImageType element to the sample DICOM file and save the file
# This is what caused the error in my real CT Scan
ds.ImageType = ['ORIGINAL', 'PRIMARY', 'LOCALIZER']


print(ds)
ds.save_as('./dicom_file_example.dcm')

client = DicomCleaner()
client.detect('./dicom_file_example.dcm')
client.clean()
```

vsoch:
okay taking a look! I'll also add the ability to load the dataset from the ds object (instead of needing to save first). I'm thinking that we had a huge factor of the DicomParser and didn't check all these functions in the cleaner. I'll know soon.

vsoch:
Here is the error I'm hitting - does this reproduce your error?

```
~/Desktop/Code/deid/deid/dicom/filter.py in empty(self, field)
    157     # This is the case of a data element
    158     if not isinstance(content, str):
--> 159         content = content.value
    160 
    161     if content == "":

AttributeError: 'NoneType' object has no attribute 'value'

```

vsoch:
This one too:

```python
[...]

PR #129, opened 2020-05-14
fixing bug with checking if list empty

vsoch:
This is a quick PR to address issue #128. I suspect that the DicomCleaner warrants a more thorough refactor to account for sequences and private tags, but this will only be possible when we have a solid real world example. This should solve the current bug that we can't parse a list of items to determine not empty. 

Related issues: # (issue)
 - #128


SpiroGanas:
Hi @vsoch 

The fix worked for me.

Thank you very much!

vsoch:
Awesome! Let's get it merged and released. Thanks for the bug report!

vsoch:
https://pypi.org/project/deid/0.2.1/

Issue #137, opened 2020-07-01
KeyError Exception when stripping sequences 


wetzelj:
When stripping sequences from DICOM headers, when a given sequence element exists in multiple locations in the header - or when sequences are nested, the ```del dicom[elem.tag]``` statement in [tags.py](https://github.com/pydicom/deid/blob/3f700388952353fc0a713a7e407b5774dd24a75c/deid/dicom/tags.py#L133) produces a KeyError.  In the CtBrain1.dcm sample file, this scenario can be seen with tag (0008, 1110) which occurs as a top-level sequence but then also occurs as a nested item within a sequence further down in the header.

I've created a test method which demonstrates this bug and implemented a two possible fixes in this [commit](https://github.com/wetzelj/deid/commit/eb4ff35a1c43465e60ce4df141a076ea2192fe35).  Let me know if which fix (if either) you prefer and if you'd like me to submit it as a pull request.

Hi @vsoch!  I'm back! :)

vsoch:
hey @wetzelj welcome back!!

I think solution A (the one not commented out) looks great - if we check for None then I don't think we will hit that key error. I vote for A, and we can adjust if there is some bug report about it.

When you open the PR (or put directly in the docstring, that's probably better) please briefly explain the test (e.g., I think we'd want to make sure that a nested value is properly removed). And I suspect we will hit other little bugs with sequences, but I'm happy to work on them as they come - smaller, well scoped PRs are good! Looking forward to your PR! Don't forget to update the changelog and do a version bump :)

wetzelj:
Closing... Fixed with PR #138

PR #138, opened 2020-07-02
Fix/remove sequences

wetzelj:
# Description

Related issues: #137 
Added additional check in tags.py to ensure that the element exists prior to attempting the element deletion in the remove_sequences function.  Nested sequences will be removed when the parent sequence is removed, and therefore, when the child is attempted to be removed (as a result of being in the initial dicom.iterall()), it will not be present, previously resulting in a KeyError. This fix, checks to ensure that the sequence being deleted still exists in the header prior to attempting the deletion.

# Checklist

# Open questions

To keep this PR tightly scoped, this is limited to just the strip_sequences functionality with replace_identifiers.  This is not intended to evaluate or test specific actions against tags that exist as top-level tags and child tags within a sequence.


wetzelj:
Hey @vsoch... Do you know how I can fix the formatting failure?  I can't figure out what it doesn't like...  I don't like failing! :)

vsoch:
Looks great! Sorry for the delayed response, GitHub pages has been coughing and I've been trying to release the RSE Stories podcast episode - it finally went through. I've already reviewed this and you added a great docstring to explain the test, so I'm good to merge!

PR #141, opened 2020-08-07
Bug fix for header adds after deletes

wetzelj:
# Description

Related issues: # 140

Corrected issue occurring when a ADD action occurs after a REMOVE action.  Previously, the REMOVE would delete the tag from the header, but the field would remain in the fields list.  This created an issue with subsequent ADD actions due to the fact that the item would be found in the fields dictionary and would be treated as a replace. After correcting this issue by removing the item from the fields list, it introduced an issue when iterating over the fields dictionary while processing multiple-field actions. To accommodate this the iteration over the fields dictionary was switched to a clone so that the dictionary on which we are iterating would not be changed by potential REMOVE actions. 

# Checklist

# Open questions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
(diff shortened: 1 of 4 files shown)


vsoch:
As usual, you've done a great job @wetzelj ! If you are done working, I'm good to merge and release!

wetzelj:
@vsoch - I'm good with it if you are!

vsoch:
Thank you! https://pypi.org/project/deid/0.2.15/

Issue #144, opened 2020-09-09
Empty values within FIELDS identified in %values sections


wetzelj:
This bug presents itself in two locations:
1. When a %values lists contains a FIELD which has an empty value, the empty string is used to search for fields on which to act in the action.  This causes all fields to be identified and all header fields to be removed. 
2. When a %values lists contains only FIELDS/SPLITS which result in empty values within the values list, the empty list is used to when searching for fields to target and results in the removal of all header fields. 

Values being added to a %values list should be interrogated so that empty strings are not added to values lists and %values lists should be interrogated so that if a values list contains no values, the action is ignored. 

I've put together a possible fix as well as unit tests for this condition within my [fix/empty_values_lists](https://github.com/wetzelj/deid/tree/fix/empty_values_lists) branch.  Let me know what you think.

wetzelj:
FYI... I recently updated to black v20.8b1.  It appears to have introduced some changes to the docstring quotes.  If there's a specific version I should be using, let me know. :)

vsoch:
Your branch looks great! One suggestion, it's slightly more python to do:

```python
if values:
```
instead of
```python
if len(values) > 0:
```

For black we install the latest version, so it might be okay (we can see how it runs when the PR is opened). We can pin a version if necessary after that. Open away! :)

vsoch:
Thank you again for this! It’s really lovely to have someone to collaborate with, and fulfilling to see the library getting better.

PR #145, opened 2020-09-09
Fix/empty values lists

wetzelj:
# Description

Related issues: #144  

Includes the fixes to ensure that empty values added to %values sections are ignored.  Prior to this fix, empty values would insert an empty string into the values list, causing all fields to be removed from the header.  This PR corrects these issues and introduces new unit tests for checking this issue.

# Checklist

# Open questions


vsoch:
Looks good to me! We could still use an iterator for going through the values list (and checking for "" or None) but I think it works well as is too. Let's get this merged and released! And thank you as always for your careful, detailed work.

Issue #147, opened 2020-09-11
Error with private tags in %fields sections


wetzelj:
Including private tags within fields sections cause all private tags to be acted upon. 

Currently, in parser.perform_action, when a fields list is encountered, [expand_field_expression](https://github.com/pydicom/deid/blob/bb98013b4dad4674af0e4f338562a6773d4f8132/deid/dicom/parser.py#L333) is called using the keyword from the contender element.  In the case of private tags, the keyword of the element is an empty string.  When this empty string is utilized by expand_field_expression to obtain the list of fields, it returns all private tags.

I believe that this can be corrected by passing the stripped_tag into expand_field_expression so that the unique tag number is utilized to identify the tag rather than the tag name - [started](https://github.com/wetzelj/deid/commit/1f5c7823348fb8d38c0d9b4e902a20472cafbd35).

wetzelj:
@vsoch - I'm about to sign off for the week.  Will it be okay if I handle both this and #146 in a single PR next week or would you prefer them split?  Have a great weekend!

vsoch:
Yes definitely okay to combine. Have a good weekend!

vsoch:
Fixed with #148.

Issue #146, opened 2020-09-11
Default VR for added private tags


wetzelj:
When discussed in #127, we decided to use DICOM VR "UN" (undefined) as the default data type for a private tag. While this seemed to be okay when tested via our current unit tests, this actually causes an exception when calling pydicom FileDataset.save_as to write the output DICOM file due to the fact that the UN datatype is expecting bytes, not a string. 
>'With tag (1111, 2221) got exception: a bytes-like object is required, not \'str\'

Would you be open to defaulting new private tags to one of the string datatypes?  

- LT - Long Text (10240 chars max)
- LO - Long String (64 chars max)
- ST - Short Text (1024 chars max)
- UT - Unlimited Text

If you think this is okay, I'll make the change and add a couple unit tests.

vsoch:
We should likely choose the one that is reasonable and has least potential to spit an error. If we were to do some kind of text, would adding a number spit out an error? Short text seems reasonable to not be too short or too long, and we'd want to discuss if it would work for other types too (or if we should check for the type in python and choose based on that).

wetzelj:
Based on what I can tell from some initial evaluation, the ST type will be accepting of most variants we would see. Its accepted positive/negative integers, decimals, and even a string with lengths greater than the limits defined in the standard. 

I don't think it's necessarily worth checking in python and dynamically setting the type.  If we could definitively determine types, then maybe - but given the way the VRs are [defined](http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html), we'd have to make some ambiguous decisions.

vsoch:
Ok, let’s go with short text then!

vsoch:
Closed with #148, type changed to short text.

PR #148, opened 2020-09-14
Bug fixes for private tags

wetzelj:
# Description

Related issues: #146 , #147 

- Modified the default VR for newly added tags from UN (Unknown) to ST (Short Text).  (#146)
- Replicated replace identifiers unit tests to ensure that modified headers can be appropriately written as files.  (#146)
- Modified expand_field_expression to utilize the tag number instead of the element keyword when processing %fields lists. (#147)

# Checklist

# Open questions


deid/tests/test_replace_identifiers_write.py
```diff
@@ -0,0 +1,892 @@
+#!/usr/bin/env python
+
+"""
```
vsoch, line 3: Two quick questions here:

 - Do you know why this is rendering as a completely new file instead of rename of https://github.com/pydicom/deid/blob/master/deid/tests/test_replace_identifiers.py?
 - Why the rename?

vsoch (commented):
Looks good! If possible, we should commit changes to the `test_replace_identifiers_write.py` so it shows up as a rename instead of entirely new file.

deid/tests/test_replace_identifiers_write.py
wetzelj: I went back and forth (and honestly, today, I'm on the other side).  Since the issue only appeared when the save=True in replace_identifiers, I thought that it could be beneficial to have a set of tests where save=True (test_replace_identifiers_write) and a set where save=False (test_replace_identifiers).  But really, this is redundant since shifting everything to write is still testing the "save=False" scenario as well.

vsoch: Can you point me to the case that had a different result for save vs not? I would suggest just testing the save_identifiers cases with each of save is True and save is False.

wetzelj: By flipping the default Tag VR to "ST", save=True and save=False do behave the same (#146).  However when the default Tag was "UN", we saw different behavior in [test_add_private_constant](https://github.com/pydicom/deid/blob/bb98013b4dad4674af0e4f338562a6773d4f8132/deid/tests/test_replace_identifiers.py#L58) and [test_add_tag_variable](https://github.com/pydicom/deid/blob/bb98013b4dad4674af0e4f338562a6773d4f8132/deid/tests/test_replace_identifiers.py#L196).

I'm good with however you think is best to test this. My initial thought yesterday was to have a save=False version of the tests (test_replace_identifiers.py) and a save=True version of the tests (test_replace_identifiers_write.py) - that's what I initially submitted with the PR - but it is redundant. If you'd prefer it, I could just simply create second versions of test_add_private_constant and test_add_tag_variable.

vsoch: Yes I like that approach - second versions of specific tests that have issue, but for the rest we probably don’t need to save to file. Does that work?

wetzelj: That's fine with me!  I'll commit it in a few.

vsoch (approved):

wetzelj:
Thanks!

vsoch:
https://pypi.org/project/deid/0.2.19/

Issue #149, opened 2020-09-16
Pixel cleaner failures


wetzelj:
There are a couple small issues in clean.py:

1. Many old recipes contain filter sections which only contain coordinates and do not contain keepcoordinates properties.  The absence of keepcoordinates causes a KeyError exception when attempting to retrieve the keepcoordinates from the DICOMCleaner results ([clean.py - line 153](https://github.com/pydicom/deid/blob/63dc9e2b43b99aae8d63b51667195d441c86e5c4/deid/dicom/pixels/clean.py#L153))

2. For 2-dimensional images, the cleaner falls into the else condition ([clean.py - line 201](https://github.com/pydicom/deid/blob/63dc9e2b43b99aae8d63b51667195d441c86e5c4/deid/dicom/pixels/clean.py#L201)) which causes no pixel masking to be performed and an exception to be thrown because self.original.shape is not converted to a string prior to being used in the format statement.

I've committed proposed fixes in [this commit](https://github.com/wetzelj/deid/commit/9258a1654a5857d9d5deda4fce4fd004f45aa40b).  Let me know if you have any suggestions on this.  It's the first I've looked at pixel cleaning functionality.

vsoch:
Thank you @wetzelj ! I added some comments on the code in case it doesn't send a notification.

wetzelj:
Fixed with #150.

PR #150, opened 2020-09-17
Updates to clean and detect

vsoch:
# Description

Opening this PR as start of work to refactor the detect and clean flow, changes include:

- removing the ultrasound regions from the default deid.dicom and adding one specifically for ultrasound
- parsing them into one list, in the order provided, so keepcoordinates (value of 1) and coordinates (0) are respected for the order the user wants
- having the default value set to ones (indicating keep)
- adding an ability to say coordinates all or keepcoordinates all in case the user wants to apply a change to the entire mask (e.g., for the ultrasound we would do coordinates all to blank them first before keeping the Ultrasound region.
- parsing the dicom in clean.py to account for these changes, and also adding image dimension of 2.
- adding documentation that explains this!

cc @wetzelj 

Related issues: #149 

# Checklist

- [ ] My code follows the style guidelines of this project

# Open questions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,7 @@ jobs:
         run: |
           export PATH="/usr/share/miniconda/bin:$PATH"
           source activate black
+          pip install black==20.8b1
           black --check --verbose deid
 
   testing:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - fixes to detect and clean to better represent keep/coordinates (0.2.20)
  - modify default VR for added tags [#146](https://github.com/pydicom/deid/issues/146), bug with private tags in %fields section [#147](https://github.com/pydicom/deid/issues/147) (0.2.19)
  - adding keepcoordinates to allow for reverse specified coordinates (0.2.18)
  - bug with empty values within %values sections [#144](https://github.com/pydicom/deid/issues/144) (0.2.17)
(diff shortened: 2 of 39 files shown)


wetzelj:
@vsoch - Are you okay with having mini, unit-test specific recipe files added for testing clean?  If so, would a new "resources" folder under tests be okay?

vsoch:
Definitely!

wetzelj:
I have some unit tests written as well as some bug fixes.  They're currently committed [here](https://github.com/wetzelj/deid/tree/pr/150_wetzelj), but I can port them to where ever you want if you are okay with the changes. 

Just to point out a couple things:
- Removed the looping on line 156 in clean.py.  
- Switched the creation of the output folders from a mkdir to makedirs so that the entire tree is created.
- Removed the "cleaned-" prefix from the output header files. This simplifies the merging of the pixel output with header output.

vsoch:
Great! Let's do the PR against the branch here, then we can merge, make sure all the tests still work, and then merge again and release. Everything sounds good except for:

> Removed the "cleaned-" prefix from the output header files. This simplifies the merging of the pixel output with header output.

The concern here is that the user would accidentally rewrite files - I think we should keep the cleaned- prefix. It's also a message that the files were absolutely generated - without the prefix it might not be clear. If you feel strongly that this isn't a good feature, then we could have some kind of input argument for an output prefix, with default to be `cleaned-` and for your use case setting it to an empty string.

wetzelj:
Can you check my permissions?  I'm getting access errors when trying to commit here.

As far as the "-cleaned" prefix goes, I don't have a strong preference here.  Let's just leave it as is - with prefix.

wetzelj:
Got it. 

I don't have a strong preference on the "cleaned-" prefix.  Let's just leave it as is.

vsoch (commented):
@wetzelj this looks good to me, ready for release?

wetzelj:
Same here.  I will be continuing to test.  If I see anything else I'll open another issue.

vsoch:
I'll give you a day or so to test and report any quick issues, otherwise I'll release after that!

PR #151, opened 2020-09-22
Pr150 - Bug Fixes and Tests

wetzelj:
# Description

Related issues: #149 

- Removed the looping on line 156 in clean.py to correct bug when adding coordinates from recipe file.
- Switched the creation of the output folders from a mkdir to makedirs so that the entire tree is created.
- Converted self.original.shape to string prior to displaying in warning message.  Previously threw an exception.
- Added unit tests for the DICOMCleaner clean functionality.

# Checklist

# Open questions


wetzelj:
I'm not planning to fix the formatting errors, as I think it would be contrary to what we actually want.  I believe these are occurring because within this PR, the CI formatting check is still checking using black 19.10b0.  The file is formatted using black 20.8b1.  I noticed you bumped black to version 20.8b1 in your last [commit](https://github.com/pydicom/deid/pull/150/commits/24049ba27ca327242b906c1257f82bfa15c9a399).  Let me know if you want me to do something with this.

deid/dicom/pixels/clean.py
```diff
@@ -153,36 +153,34 @@ def clean(self, fix_interpretation=True, pixel_data_attribute="PixelData"):
...
-                            if len(self.original.shape) >= 3:
-                                new_coordinate = [
-                                    0,
-                                    0,
-                                    self.original.shape[2],
-                                    self.original.shape[1],
-                                ]
-
+                    # Case 1: an "all" indicates applying to entire image
+                    if new_coordinates.lower() == "all":
```
vsoch, line 157: This should be a list of coordinates, e.g., 
```
for new_coordinate in new_coordinates
```

deid/dicom/pixels/clean.py
```diff
@@ -153,36 +153,34 @@ def clean(self, fix_interpretation=True, pixel_data_attribute="PixelData"):
...
+
+                        # no frames, just X, Y
+                        if len(self.original.shape) == 2:
+                            # minr, minc, maxr, maxc = [0, 0, Y, X]
+                            new_coordinates = [
```
vsoch, line 162: And then this is the singular of that, new_coordinate (setting new_coordinates here would overwrite the original variable. The reason that we allow a rule to have more than one coordinate is because there might be more than one relevant region.

vsoch (commented):
This is beautiful work @wetzelj ! And I'm okay with merging in without formatting to my branch - I can pull and format once we merge. Before we merge here, could you walk me through why you removed the list iteration from parsing the coordinates? Each coordinate set should be a list, so then all the coordinates for a rule should be a list of lists. For the most part that means that single sets of coordinates look like:

```
[[0,0,0,0]]
```
but of course we could have more than one
```
[[0,0,0,0], [1,1,1,1]]
```

deid/dicom/pixels/clean.py
wetzelj: Is this a difference in how identified coordinates come in through a recipe vs the "from:field" functionality?  If we look at base [deid.dicom](https://github.com/pydicom/deid/blob/master/deid/data/deid.dicom#L297) recipe, instances of multiple regions being detected by a filter are identified by multiple coordinates lines.  Each one of these coordinates lines results in a separate "coordinate set".   With the loop in pace, this iterates through a single coordinate comma separated string and not through a list of coordinates.

Two of the test cases that I wrote specifically test for the multiple-region scenario, however, they are all recipe based tests.  I may need the sample ultrasound you mentioned to write some more tests if there's a difference in the way the coordinates are identified through the "from:Field" functionality

deid/dicom/pixels/clean.py
wetzelj: Absolutely. This is intentional - see my comment above.  With this overwriting we are still allowing for multiple coordinates - when specified as in the deid.dicom recipe.

wetzelj:
I couldn't let the formatting go. :)

deid/dicom/pixels/clean.py
vsoch: There isn't a difference - whether the coordinate comes in via a from or a list, from, or all, it should still be rendered as a list, eg.,
```
['from:FieldName', 'all', '0,0,0,0']
```
So a recipe should be able to have more than one coordinate, possibly even in different formats. I'll send you the echo image right away!

vsoch:
Haha ok :) This LGTM - are you ready to merge into the other branch and have everything run together?

wetzelj:
As you've probably noticed, the loop is back in - I just had to introduce a little check to ensure we had a list.  When coordinates are identified from recipe files, they were just returned as a string - not in a list.  

Actually, thinking about it a bit more, do you think I should change detect.py ([line 208](https://github.com/pydicom/deid/blob/24049ba27ca327242b906c1257f82bfa15c9a399/deid/dicom/pixels/detect.py#L208)) to return the coordinate from recipe as a list?  I think regardless we'll want to retain the list check I introduced as a line of defense.

wetzelj:
Otherwise, if you're good with it as is, I'd say we're good to merge.

vsoch:
I think it was working as is - let's leave for now and change if some future discovery of a bug warrants it.

Issue #142, opened 2020-08-31
Incorrect logic in detect.py module


rmw362:
Hi,

I really appreciate all of your efforts on this project...super helpful. I came across some behavior of the detect.py module that I believe is counter to how you intended the module to work. 

First, the logic of how the detect.py module flags images is incorrect, and this is due to improper indentation in the evaluate groups function starting with "#At the end of the group, evaluate the inner group" below. This section needs to be indented to be included with the preceeding for loop (as I have done below), otherwise it is evaluating outer groups rather than inner groups (this is the current functionality as written). The result is inappropriate behavior where if the header meets the final criteria in the outer groups, it will be flagged regardless of whether it met criteria for previous groups. For example, I have an US DICOM which has 768 rows in the pixel data, and the image was flagged with every filter with equals Rows 768 as the last criteria. Thankfully a simple tab solves the issue. Additionally, you need to rename the output as overall_result to avoid a namespace conflict with results so that this solution works. 

```
def _has_burned_pixels_single(dicom_file, force, deid):

    """has burned pixels single will evaluate one dicom file for burned in
       pixels based on 'filter' criteria in a deid. If deid is not provided,
       will use application default. The method proceeds as follows:
       1. deid is loaded, with criteria groups ordered from specific --> general
       2. image is run down the criteria, stops when hits and reports FLAG
       3. passing through the entire list gives status of pass
    
       The default deid has a greylist, whitelist, then blacklist
       Parameters
       =========
       dicom_file: the fullpath to the file to evaluate
       force: force reading of a potentially erroneous file
       deid: the full path to a deid specification. if not defined, only default used
       deid['filter']['dangerouscookie'] <-- filter list "dangerouscookie"
       --> This is what an item in the criteria looks like
            [{'coordinates': ['0,0,512,110'],
              'filters': [{'InnerOperators': [],
              'action': ['notequals'],
              'field': ['OperatorsName'],
              'operator': 'and',
              'value': ['bold bread']}],
            'name': 'criteria for dangerous cookie'}]
    
       Returns
       =======
       --> This is what a clean image looks like:
           {'flagged': False, 'results': []}
       --> This is what a flagged image looks like:
          {'flagged': True,
           'results': [
                  {'reason': ' ImageType missing  or ImageType empty ',
                   'group': 'blacklist',
                   'coordinates': []}
               ]
           }
    """
    dicom = read_file(dicom_file, force=force)

    # Load criteria (actions) for flagging
    filters = deid.get_filters()
[...]

PR #152, opened 2020-09-30
Fix/compound filters

wetzelj:
# Description

Related issues: #142 

Corrected a bug which caused only the last line of a defined filter to be utilized when using the DICOMCleaner to detect PHI in images.  Added unit test to validate various forms of adding filter rules into the recipe file. 

# Checklist

# Open questions

### DIFF
--- a/deid/dicom/pixels/detect.py
+++ b/deid/dicom/pixels/detect.py
@@ -142,9 +142,9 @@ def _has_burned_pixels_single(dicom_file, force, deid):
                 group_descriptions = [item.get("name", "")]
 
             else:
+                group_flags = []  # evaluation for a single line
+                group_descriptions = []
                 for group in item["filters"]:
-                    group_flags = []  # evaluation for a single line
-                    group_descriptions = []
 
                     # You cannot pop from the list
                     for a in range(len(group["action"])):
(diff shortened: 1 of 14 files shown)


deid/dicom/pixels/detect.py
```diff
@@ -142,9 +142,9 @@ def _has_burned_pixels_single(dicom_file, force, deid):
                 group_descriptions = [item.get("name", "")]
 
             else:
+                group_flags = []  # evaluation for a single line
```
vsoch, line 145: I think what probably happened here, because it's such an obvious mistake, is that I originally had a set of flags per item, but decided to change it to be on the level of the group, and didn't move the definition outside of the loop as I should. Aside from that, no idea why I did that, maybe just tiredness and then lack of testing it.

vsoch (approved):

vsoch:
https://pypi.org/project/deid/0.2.21/

deid/dicom/pixels/detect.py
wetzelj: It happens to the best of us... and now, we have some test methods for it! 😃

vsoch: You're the best to work with @wetzelj ! Your work is always so thorough, not just with the fix, but tests too. You're by far the best random collaboration I've found on here! Maybe some day if there is some project that warrants it we can write a software paper proper for it. Or since you've been so helpful, I'd be happy to do most of the writing. Cheers!

wetzelj: You've been great to work with too @vsoch!  This has been a fun effort.  At some point, I'll share how we're utilizing deid. 
We're still in the initial development of our end-user application, but are close to having something to be able to start end-user testing! 

Have a great evening!

Issue #155, opened 2020-10-02
Detect exits when recipe does not contain filters


mjcbello:
Hi @vsoch ,

I've been working with @wetzelj on our internal deid pipeline project. It has been really interesting working with all things DICOM and pydicom/deid. Thank you for your contributions!

During my testing, I noticed that calling detect on the DicomCleaner results in a bot.exit() if the current recipe file does not contain any filter sections. Is this the intended behavior of the application? Any objections to returning flagged=False instead of sys.exit?

detect.py lines 124-127:
```
    # Load criteria (actions) for flagging
    filters = deid.get_filters()
    if not filters:
        bot.exit("Deid provided does not have %filter, exiting.")
```

vsoch:
Definitely not! It would be good to, in addition to adding the return, print a clear message for the user (debug or info, up to you!)

mjcbello:
Sounds good! I have a fix ready. Would you like me to submit a pull request against the currently opened pull request #154 or a new pull request against master?

vsoch:
It would be easier to package it in the same version as #154 - and if you have bandwidth and are able, a review of the PR would be helpful! Sometimes users open issues and then there is a fix but they never come back to review, even if things are ready to go. So your review would be greatly appreciated!

vsoch:
And wow, I'm honored you opened your first issue here! Open source is a lot of fun, welcome to GitHub! 

![image](https://user-images.githubusercontent.com/814322/94947321-b5698c00-049a-11eb-919d-3be33fb5275d.png)

PR #156, opened 2020-10-02
Fix/155 change to return results from detect when recipe does not contain filters

mjcbello:
# Description

Related issues: #155 

Change to return results from detect when recipe does not contain filters.

# Checklist

# Open questions


wetzelj:
@vsoch  - I have no clue what we did to include all of those remote-tracking merges... Let us know if we should do something different.

vsoch:
When we merge PRs, I believe we squash and merge. So likely what you did here is rebase off of your master (with the older commits) - the changes you see are represented in master here, but since the commits are not, they appear here. If we squash and merge it should get rid of them all and just add one commit (and I'll ensure that the message doesn't include the old ones). And next time what you would want to do is generally always have your master in sync with deid master, meaning to never do a PR from your master to deid.

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
- - changing iteration technique through fields to properly add nested uids [#153](https://github.com/pydicom/deid/issues/153) (0.2.22)
+ - changing iteration technique through fields to properly add nested uids [#153](https://github.com/pydicom/deid/issues/153), change to return results from detect when recipe does not contain filters [#155](https://github.com/pydicom/deid/issues/155) (0.2.22)
```
vsoch, line 17:
```suggestion
 - changing iteration technique through fields to properly add nested uids [#153](https://github.com/pydicom/deid/issues/153) (0.2.22) 
 - change to return results from detect when recipe does not contain filters [#155](https://github.com/pydicom/deid/issues/155)
```
These can generally be still on two lines, just with the version tag at the top of the list.

deid/dicom/pixels/detect.py
```diff
@@ -124,7 +124,9 @@ def _has_burned_pixels_single(dicom_file, force, deid):
     # Load criteria (actions) for flagging
     filters = deid.get_filters()
     if not filters:
-        bot.exit("Deid provided does not have %filter, exiting.")
+        bot.warning("Deid provided does not have %filter.")
+        results = {"flagged": False, "results": None}
```
vsoch, line 128: What we would want to do here is move the lines under this block to be above it, e.g., these two:

https://github.com/pydicom/deid/blob/5bae2a7725dd845e24dec17e063430eec11c0470/deid/dicom/pixels/detect.py#L132

And then instead of returning None for results (will trigger an error if the user is expecting a list to loop through) and hard coding False, just do:

```python
return {"flagged": global_flagged, "results": results}
```

vsoch: Perfecto!

vsoch (approved):

Issue #163, opened 2020-10-16
"stop_before_pixels" argument for "get_identifiers"


allanhummer:
Sometimes image information of the DICOM is not needed when processing the headers (e.g. for grouping multiple measurements of the same patient). Especially when dealing with large quantities of DICOMs, it would be very useful to omit the pixel array to save a lot of memory. 

read_file (dcmread), which is used by deid already has a "stop_before_pixels" argument. It would be great to access this directly when using "get_identifiers".

Thanks for providing this helpful toolbox!

vsoch:
Sure, that sounds easy to do! Here is a PR https://github.com/pydicom/deid/pull/164 for you to test, and I've added some questions I'm interested in your feedback on.

PR #164, opened 2020-10-16
adding stop_before_pixels to read_file of dicom parser

vsoch:

# Description

Adding stop_before_pixels to the Parser, to be used by get_identifiers

Related issues: #163 

# Open questions

 - is this supported across versions of pydicom?
 - are there other read_file parameters that might want to be passed, so we should ask for a dictionary of arguments instead?
 - are there other instances of read_file that a user might want this added to?
 - should the default be False?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - adding stop_before_pixels to read_file in parser (0.2.24)
  - removing verbosity of debug logger (0.2.23)
  - changing iteration technique through fields to properly add nested uids [#153](https://github.com/pydicom/deid/issues/153) (0.2.22)
  - change to return results from detect when recipe does not contain filters [#155](https://github.com/pydicom/deid/issues/155)
--- a/deid/dicom/header.py
+++ b/deid/dicom/header.py
@@ -44,7 +44,12 @@ here = os.path.dirname(os.path.abspath(__file__))
 
 
 def get_identifiers(
-    dicom_files, force=True, config=None, strip_sequences=False, remove_private=False
+    dicom_files,
+    force=True,
+    config=None,
+    strip_sequences=False,
+    remove_private=False,
+    stop_before_pixels=False,
 ):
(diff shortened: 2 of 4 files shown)


allanhummer:
Thanks for picking this up!

A dictionary of arguments would certainly be a more flexible method. I also think that other arguments of dcmread like "defer_size" or "specific_tags" could be useful.

I think False should be the default, as otherwise the image information would be missing if one does a simple load -> edit header -> save procedure. 

Not really part of core DEID, but an instance where the "stop_before_pixels" argument is also useful is here: [create-dicom-csv.py](https://github.com/pydicom/deid/blob/master/examples/dicom/dicom-extract/create-dicom-csv.py)
Duplicate of 
Duplicate of ##

vsoch:
@allanhummer two possible suggestions for this PR:

1. add the flag to the example you provided
2. Instead of writing out specific args,have a dict for pydicom_opts (or similar) and expand it into the command. This would be flexible to any set of arguments / change over time for pydicom.

Let me know your thoughts!

vsoch:
@wetzelj what are your thoughts for this PR here? I want to make sure it doesn't get too dated to make integrating hard in case you think it's useful too. But if nobody needs it, probably okay to just let be until someone asks again :)

wetzelj:
Hey @vsoch- On the surface, I think it's a very logical inclusion.  If you don't need the pixel data, why read it?  

However, in my implementation, I'm not utilizing get_identifiers at all - I only use replace_identifiers.  Initially, I was thinking that maybe it would be good to expose the stop_before_pixels to replace identifiers as well, but ultimately, (for my purposes) it wouldn't make sense.  If we were to run with stop_before_pixels=True, since parser.dicom is what gets saved as the replaced dicom file, we'd end up with a whole mess of files that were header only, no actual images. 

While it won't be immediately beneficial to me, I think it would be a fine inclusion as it's currently coded in the pull request as an additional parameter on get_identifiers(), but honestly, I could go either way... merge the PR or let it sit until someone needs it.

vsoch:
> merge the PR or let it sit until someone needs it.

Yeah, I'm good to do the latter - keep here for discussion until someone needs it. Thanks @wetzelj !

opennog:
Hi I have a question here. 
Are you aware if the reading time when ´stop_before_pixels=True´ is quicker than when `stop_before_pixels=False`?

Thanks!

vsoch:
I am not, but you could try it out!

Issue #174, opened 2021-06-28
Adding Multivalued Fields to %values Lists


wetzelj:
When fields specified within recipe file %values lists contain a list of items, the addition of the Multi-value value to the values array causes an exception (Exception: unhashable type: 'MultiValue').  

I believe that this can be corrected checking the type of the value first - if it is [MultiValue](https://pydicom.github.io/pydicom/stable/reference/generated/pydicom.multival.MultiValue.html#pydicom.multival.MultiValue), we can use set.Update() to update the values list instead of using set.Add().  

I've committed a possible fix to the [wetzelj/deid INFOR-831](https://github.com/wetzelj/deid/tree/fix/INFOR-831) branch.

vsoch:
@wetzelj let's take these one by one! If there are easy fixes (like the one) let's do a PR from your branch and bumping the version. More complex issues (or ones you want my help with) I could take a look in an evening or weekend.

wetzelj:
@vsoch Sounds good... it's on its way

vsoch:
This is all set.

PR #176, opened 2021-06-29
Fix - Multivalued fields in %values lists

wetzelj:
# Description

Related issues: #174 

Checked value before attempting to add it to values list.  If the value is MultiValue, use set.Update() instead of set.Add().

# Checklist

# Open questions

None

### DIFF
--- a/deid/dicom/groups.py
+++ b/deid/dicom/groups.py
@@ -56,7 +56,7 @@ def extract_values_list(dicom, actions, fields=None):
         if action["action"] == "FIELD":
             for uid, field in subset.items():
                 if field.element.value not in ["", None]:
-                    if type(field.element.value) is MultiValue:
+                    if isinstance(field.element.value, MultiValue):
                         values.update(field.element.value)
                     else:
                         values.add(field.element.value)
--- a/deid/dicom/groups.py
+++ b/deid/dicom/groups.py
@@ -28,6 +28,7 @@ SOFTWARE.
 from deid.logger import bot
 from .tags import remove_sequences
 from .fields import get_fields, expand_field_expression
+from pydicom.multival import MultiValue
 
 import os
 
@@ -55,7 +56,10 @@ def extract_values_list(dicom, actions, fields=None):
         if action["action"] == "FIELD":
             for uid, field in subset.items():
                 if field.element.value not in ["", None]:
-                    values.add(field.element.value)
+                    if type(field.element.value) is MultiValue:
+                        values.update(field.element.value)
+                    else:
+                        values.add(field.element.value)
 
         # Split action, can optionally have a "by" and/or minlength parameter
         elif action["action"] == "SPLIT":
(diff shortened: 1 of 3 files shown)


deid/dicom/groups.py
```diff
@@ -55,7 +56,10 @@ def extract_values_list(dicom, actions, fields=None):
         if action["action"] == "FIELD":
             for uid, field in subset.items():
                 if field.element.value not in ["", None]:
-                    values.add(field.element.value)
+                    if type(field.element.value) is MultiValue:
```
vsoch, line 59: Let's use isinstance here, it's a bit faster: https://switowski.com/blog/type-vs-isinstance
```suggestion
                    if isinstance(field.element.value, MultiValue):
```

deid/dicom/groups.py
```diff
@@ -55,7 +56,10 @@ def extract_values_list(dicom, actions, fields=None):
...
+                        values.update(field.element.value)
```
vsoch, line 60: So just so I understand - when we have a MultiValue it is a dictionary?

vsoch (commented):
This looks great - a few quick comments / questions! Is there an easy test case for this? You do a fantastic job writing tests so if the answer is no, I'm okay with taking your word that you tested this on some real data and it fixed the bug.

deid/dicom/groups.py
wetzelj: When the value is [MultiValue](https://pydicom.github.io/pydicom/dev/reference/generated/pydicom.multival.MultiValue.html), the value in field.element.value is a "list of items that are all the same type", not a single element that we're adding into the values set.   

In the specific scenario we encountered, this was with the OtherPatientIDs field.  In all but two studies that we've deidentified using pydicom/deid, the OtherPatientIDs contained a single patient identifier, however with the two outlier studies, the modality actually sent multiple patient identifiers within this value - the MultiValue value was a list of strings.

wetzelj:
> This looks great - a few quick comments / questions! Is there an easy test case for this? You do a fantastic job writing tests so if the answer is no, I'm okay with taking your word that you tested this on some real data and it fixed the bug.

I didn't create a test case for this because I would have had to fabricate multi-occurrence within the test dicom image and given the nature of this change, questioned if it was really worth it.  If you want me to, I can come up with something for it...

deid/dicom/groups.py
vsoch: Ahh gotcha, so you are essentially doing:

```python
testset= set()
testset.add('value1')
testset.update(['value2', 'value3'])
{'value1', 'value2', 'value3'}
```

vsoch:
Nah I think we are good! If you are going to do some other PRs soon then I'll merge here (but not release) and we can release an updated version when all these changes are in. As always, thanks you @wetzelj !

deid/dicom/groups.py
wetzelj: Exactly.

vsoch (approved):

Issue #173, opened 2021-06-28
Recipe Error - REPLACE Rule After JITTER Does Not Replace


wetzelj:
When testing the ability to use variable fields in recipes, we discovered that using the REPLACE keyword in a recipe for values did not work.  Both samples below should yield the same results to the file header.

**Failed:**
```
JITTER endsWith:Date -5814
REPLACE StudyDate var:myOtherDate
```
**Successful:**
```
JITTER endsWith:Date -5814
REMOVE StudyDate
ADD StudyDate var:myOtherDate
```
 
The real issue here:  When a REPLACE rule is executed following a previous JITTER rule, the replace is not executed.  When a REPLACE rule is executed independently, it does function appropriately.  

Failing unit tests have been committed to the [wetzelj/deid INFOR-646](https://github.com/wetzelj/deid/tree/fix/INFOR-646) branch to demonstrate this issue.  I have not yet done investigation of cause/fix.

vsoch:
@wetzelj I can take a look this weekend if you don't get to it.

wetzelj:
@vsoch - I looked into this a little bit today and I think I'm going to need your help on this one whenever you have a few moments.  I can see what's going on, but I don't understand it.

In a successful REPLACE action (I was using test_replace_with_constant to see this), when the value is replaced in parser.py - add_field() - on [line 397](https://github.com/pydicom/deid/blob/5f03d9f097ad86524c7afb28fbdc10cf1961d62c/deid/dicom/parser.py#L397), the statement updates the value of the item in self.fields as well as self.dicom.

Before executing line 397:
![image](https://user-images.githubusercontent.com/32198293/123868956-62741680-d8fe-11eb-8bcb-7934f099f4d4.png)
After executing line 397:
![image](https://user-images.githubusercontent.com/32198293/123869029-7881d700-d8fe-11eb-96c4-0ed2d5d80ea4.png)


In the error scenario (using test_jitter_replace_compounding), when the value is replaced in parser.py, this same line of code updates the value in self.fields, but fails to update self.dicom.
Before executing line 397:
![image](https://user-images.githubusercontent.com/32198293/123869367-e0d0b880-d8fe-11eb-874f-70122fa88b8d.png)
After executing line 397:
![image](https://user-images.githubusercontent.com/32198293/123869429-f8a83c80-d8fe-11eb-9699-2ed3505f7464.png)

[...]

PR #178, opened 2021-06-30
Fix/infor 646

vsoch:
# Description

Related issues: # https://github.com/pydicom/deid/issues/173

This will hopefully be a fix to the issue above! The bug is that we have two places to store the fields and values - the self.dicom and self.fields. The self.fields are operated on always to get and update values, but in this case we were updating fields but not the self.dicom, given that the tag was already there. I also ran black (and removed the version pin).

### DIFF
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -427,14 +427,23 @@ class DicomParser:
             )
             if value is not None:
 
-                # Jitter the field by the supplied value
+                # Cut out early if the field isn't in the dicom
+                if field.name not in self.dicom:
+                    return
+
+                # Preserve the old value in case we need to update
                 old_value = self.dicom[field.name].value
+
+                # Jitter the field by the supplied value
                 jitter_timestamp(
                     dicom=self.dicom, field=field.element.keyword, value=value
                 )
+
+                # Get the updated value, update if it's different
                 new_value = self.dicom[field.name].value
                 if old_value != new_value:
                     self.replace_field(field, new_value)
+
             else:
                 bot.warning("JITTER %s unsuccessful" % field)
 
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -257,9 +257,7 @@ class DicomParser:
         """
         if not self.fields:
             self.fields = get_fields(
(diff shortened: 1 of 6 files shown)


deid/dicom/parser.py
```diff
@@ -429,9 +428,13 @@ def _run_action(self, field, action, value=None):
             if value is not None:
 
                 # Jitter the field by the supplied value
+                old_value = self.dicom[field.name].value
```
vsoch, line 431: I didn't check if it might be the case the value doesn't exist in the dicom.

vsoch:
I don't think I have a perfect answer, but I can say that the bug resulted from having two ways of doing this (the older one with using the dicom and then the updated one with the parser and saving fields) that (as we discussed and you pointed out) weren't consistent. The two areas that stuck out to me as causing this specific problem were here: https://github.com/pydicom/deid/blob/ad37ee4cc8339bd3bdebabca33274e4cc63fe817/deid/dicom/parser.py#L390

A call to replace actually calls add_field, 
```python
                if uid in self.fields:
                    element = self.fields[uid]

                    # Nested fields
                    while not hasattr(element, "value"):
                        element = element.element
                    element.value = value

                else:
                    element = DataElement(tag["tag"], tag["VR"], value)
                    self.dicom.add(element)
                    self.fields[uid] = DicomField(element, name, uid)
```
and you'll notice if the uid is already in fields, we just update the value, but we don't modify the dicom. The top loop is missing an update to self.dicom, which I added here. I think the assumption was that the element updated points to the same object in the dicom? The jitter timestamp was the opposite - we made changes to self.dicom without updating fields, here: https://github.com/pydicom/deid/blob/ad37ee4cc8339bd3bdebabca33274e4cc63fe817/deid/dicom/parser.py#L432. Again, likely the assumption being that the update to self.dicom also updates the object in self.fields (which is the same reference, at least if I remember a long time ago from testing). 

I don't really have a good answer for why using JITTER seemed to break things, but I suspect there was something out of sync with respect to the two places (self.dicom and self.fields). I'd have to debug it again to check, but if I remember correctly we wound up with the result of the jitter instead of the result of the final replace.

Issue #175, opened 2021-06-28
Date Jitter on %values Lists


wetzelj:
During the deidentification of a study a situation was encountered where the StudyDate value existed within a private dicom tag.  This tag (0029,1019) was of data type LO, but contained the StudyDate value:

```(0029,1019) : LO   Len: 8 <Unknown Tag> Value: [20100526]```
We attempted to handle this using current recipe functionality:
```
%values testing 
FIELD StudyDate 
%header 
JITTER values:testing -3241
```
However, performing these actions did not jitter the value in the Private tag.
 
**Root Cause**
pydicom/deid (deid\dicom\actions.py - jitter_timestamp())
This function expressly performs JITTER only if the DICOM VR is DT or DA.
 
**Idea**
Figure out how to handle non-date typed values.  We could attempt to parse the string representations of the [DA/DT types](http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html), and if it doesn't produce an exception, jitter it.

vsoch:
This is closed by #181

PR #180, opened 2021-06-30
Allow other vr types for jitter

vsoch:
# Description

@wetzelj I'm not sure if this is what you had in mind? The idea is that if the type doesn't match, we don't know the format string so we try them all until we find one that works.

Related issues: #175 

I'm not sure if we need to parse the field differently (e.g., if it's a list). Could you show me an example if this is the case? I also accidentally ran black because it's sort of second nature, sorry for the extra changes to parse through!


wetzelj:
Hey @vsoch - I'm running out of time for the day and I didn't want to leave this hanging, but don't have anything ready to cleanly add to the pull request. I've committed adjustments to the ctbrain1.dcm file as well as a new unit test to a new branch on my fork ([other-vr-jitter](https://github.com/wetzelj/deid/tree/other-vr-jitter)).  The unit test currently fails, but its due to the fact that the field that I placed the "other VR date" in is a private tag - and jitter_timestamp isn't happy with it, when jitter_timestamp is called, field.element.keyword is passed in as the field - and it's blank for the private field.

What you've implemented is definitely what I had in mind. 

Have a great weekend!

vsoch:
Ah, no worries! We can pick this up when you have time. Does the other PR look good (and would you like me to merge and release, or wait for this fix too?)

wetzelj:
The other PR looks good, but if you don't mind, let's hold the release for this one too... I'll get back to it early next week.

wetzelj:
Hey @vsoch - Check out the latest commit on my fork ([other-vr-jitter](https://github.com/wetzelj/deid/tree/other-vr-jitter)).  In building the unit test for this, I discovered that Jitter really didn't work for private tags since the jitter functionality was relying on the keyword to get the field. I went a bit out on a limb fixing this - thought that really it would be better to update the dicom/fields in one place, so I refactored jitter_timestamp to just return the new value, so that we could then just use replace_field() to set the new value.  If you don't like this approach, I'm totally fine scrapping it, but if you do think it's okay, I'll submit a PR.

vsoch:
@wetzelj yes this is so much better! I was thinking about how the design for this function is different from all others (e.g., it interacts with the dicom directly) and that's an artifact of how deid was originally designed. If you want to PR with this update I'll review asap and totally okay to replace this branch of work.

vsoch:
Superseded by #181

wetzelj:
@vsoch  - I think this needed to be merged into master - not closed.  #181 was based on what you started here.

vsoch:
oh my mistake - I thought you had included these commits?

vsoch:
@wetzelj could you look at why the test is failing? I think you made some updates that I'm not privy to, I wasn't expecting it to fail!

wetzelj:
Yep.  Looking into it.

wetzelj:
This was all my bad.  Sorry.

There were some conflicts between the fix for #175 and #178.  Since I hadn't merged the fix for #178 in, when you merged to master, the conflicts caused the error. 

Again... sorry for the issues.

vsoch:
No need to apologize - you really didn't do anything wrong! These kinds of things happen with multiple changes going on at once. I'm very grateful that you fixed everything up, and we now have a release!

PR #181, opened 2021-07-07
Addition to PR#180 - new unit tests, jitter_timestamp changes

wetzelj:
Addition to PR#180 - new unit tests, jitter_timestamp changes

# Description

Related issues: #175 
Added unit tests to validate jittering of private tag fields as well as jitter of multiple fields identified by a values list.   Refactored jitter_timestamp to return the new jittered value and use replace_field to actually perform the modification of the field value. 

# Checklist

# Open questions


deid/dicom/actions.py
```diff
@@ -23,47 +23,36 @@
...
-    dicom: the pydicom Dataset
     field: the field with the timestamp
     value: number of days to jitter by. Jitter bug!
 
     """
     if not isinstance(value, int):
         value = int(value)
 
-    original = dicom.get(field)
+    original = field.element.value
```
vsoch, line 45: Could this be none first? (field.element)

deid/dicom/actions.py
```diff
@@ -23,47 +23,36 @@
...
+    new_value = original
 
     if original is not None:
 
         # Create default for new value
         new_value = None
-
-        # If we have a string, we need to create a DataElement
```
vsoch, line 57: hmm I think since we just pass field as is, it could still be a string?

deid/dicom/parser.py
```diff
@@ -427,9 +427,8 @@ def _run_action(self, field, action, value=None):
             if value is not None:
 
                 # Jitter the field by the supplied value
-                jitter_timestamp(
-                    dicom=self.dicom, field=field.element.keyword, value=value
-                )
+                new_val = jitter_timestamp(field=field, value=value)
```
vsoch, line 430: Re my comment above: could field ever be a string here?

deid/dicom/actions.py
wetzelj: I don't think so.  I checked the places that create DicomField objects, and they're checking to ensure that element is a DataElement prior to the creation of the DicomField.

deid/dicom/actions.py
wetzelj: I don't believe so.  By the time we're performing this (via _run_action()), the string representation of the field has been evaluated to be a DicomField object by find_by_values() or expand_field_expression().   When parser.perform_action() calls _run_action(), we're iterating over a list of DicomFields.

deid/dicom/parser.py
wetzelj: See above.

vsoch (approved):

PR #185, opened 2021-09-08
Typos found by codespell

DimitriPapadopoulos:
# Description

Fix typos detected by [codespell](https://github.com/codespell-project/codespell).

# Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [ ] My code follows the style guidelines of this project

PR #187, opened 2021-09-15
Add codespell to CI to catch typos early

DimitriPapadopoulos:
# Description

Run [codespell](https://github.com/codespell-project/codespell) in CI to catch typos early on.

Use GitHub Actiosn with [actions-codespell](https://github.com/codespell-project/actions-codespell).

Related PR: #185

# Checklist

- [ ] My code follows the style guidelines of this project

### DIFF
--- a/.github/codespell_ignore_words.txt
+++ b/.github/codespell_ignore_words.txt
@@ -1,0 +1,4 @@
+nd
+ot
+searchin
+som
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,17 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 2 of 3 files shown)


vsoch:
Good idea! Please use an exact sha instead of a branch or tag so it's not a moving target.

DimitriPapadopoulos:
For what it's worth, it's supposed to be used like this:
```
uses: codespell-project/actions-codespell@master
```
See https://github.com/codespell-project/actions-codespell#usage

DimitriPapadopoulos:
But I agree there is a risk.

vsoch:
Yes I'm aware how GitHub actions work - for actions that I don't control my preference is to use a non moving target. A commit sha, and not a branch or release tag.

DimitriPapadopoulos:
OK, I have used the current commit.

vsoch:
Perfect! I think you just need to provide the entire sha and it will work.

vsoch:
Thanks! This will prevent my future misspellings :)

DimitriPapadopoulos:
In the future, false positives can be added to **`.github/codespell_ignore_words.txt`**. Hopefully there won't be many.

PR #191, opened 2021-10-08
Fix LGTM recommendations

DimitriPapadopoulos:
# Description

Fixes most of the remaining LGTM recommendations:
https://lgtm.com/projects/g/pydicom/deid/alerts/?severity=recommendation

# Checklist

# Open questions

The remaining recommendations should probably not be fixed.

### DIFF
--- a/deid/dicom/pixels/clean.py
+++ b/deid/dicom/pixels/clean.py
@@ -118,7 +118,6 @@ class DicomCleaner:
             # Get expected and actual length of the pixel data (bytes, expected does not include trailing null byte)
             expected_length = get_expected_length(dicom)
             actual_length = len(pixel_data)
-            padded_expected_length = expected_length + expected_length % 2
             full_length = expected_length / 2 * 3  # upsampled data is a third larger
             full_length += (
                 1 if full_length % 2 else 0


DimitriPapadopoulos:
About [Except block handles 'BaseException'](https://lgtm.com/rules/6780080/):

> All exception classes in Python derive from `BaseException`. `BaseException` has three important subclasses, `Exception` from which all errors and normal exceptions derive, `KeyboardInterrupt` which is raised when the user interrupts the program from the keyboard and `SystemExit` which is raised by the `sys.exit()` function to terminate the program.
> 
> Since `KeyboardInterrupt` and `SystemExit` are special they should not be grouped together with other `Exception` classes.
> 
> Catching `BaseException`, rather than its subclasses may prevent proper handling of `KeyboardInterrupt` or `SystemExit`. It is easy to catch `BaseException` accidentally as it is caught implicitly by an empty `except`: statement.

DimitriPapadopoulos:
I must have misunderstood you. I believe you want automatic LGTM reporting (not automatically fixing alerts). Of course it is possible to have a CI script that checks new PRs and raises alerts. Here is the relevant page:

[Integration with GitHub Apps](https://lgtm.com/help/lgtm/github-apps-integration)

vsoch:
Thanks for the details! I will have to think about it - I'm usually pretty conservative about giving bots write access to PRs (or anything for that matter) and I'm not sure the benefits outweigh the risk. These are pretty small, trivial things.

image

PR #195, opened 2021-11-09
Fix `Identifiction` -> `Identification` typo

agirault:


vsoch:
Oh thank you! There is a spell checking action that I like (and use on some of my other repos) we should probably add it here - typos are unavoidable. :P

agirault:
Right, not a big deal, but that jumped to me since that was in the title of the main webpage.

I did wonder for a second if that wasn't on purpose (isn't perfect de-identification just **fiction**? **de-identi-fiction**!), which would not be that crazy given the reasonable claims to only provide **best effort** anonymization. 😄

vsoch:
haha, that's pretty funny :)

I will have a PR this evening to add another spell checker - we have code spell but it's missing the examples folder. Thanks for helping!

Issue #197, opened 2021-12-07
action order on same field


glebsts:
Hello,
I am confused by following in [docs](https://pydicom.github.io/deid/user-docs/recipe-headers):

> For the above, given that there are conflicting commands, the more conservative is given preference. For example:
> REMOVE > BLANK > REPLACE > JITTER/KEEP/ADD
> ...
> If I specify to blank a header and remove it, it will be removed
> If I specify to replace a header and blank it, it will be blanked

Also

> Most of the time, you won’t need to specify remove, because it is the default

But at the same time there is an example of
```
ADD PatientIdentityRemoved Yes
REMOVE ALL
KEEP PixelData
```
Here I naturally make assumption that `ALL` means `ALL`, so each field is removed, and according to rules quoted above, `REMOVE` rules over `KEEP`, so tag kinda has to be removed.
How does it work? How is `PixelData` expected to be kept?

In my usecase I want to drop all information except for certain tags, part of them I want to keep, part of them I want to replace with my custom func.
I don't want to manage full set of DICOM tags, but rather remove all and preserve only certain subset.
How?

vsoch:
I would try to follow the example and do a REMOVE ALL and then add the fields you want to keep. If that is successful we can update the docs to reflect it - I think the spirit of the docs is that if you were to have all those commands in the same file for the same field, it's saying that `REMOVE` would take preference. But it also could be that they are out of date.

Please test it out for your use case and report back!

glebsts:
I tested on 
```
REMOVE ALL
KEEP SeriesDescription
```
and it ends up with everything removed. Same if I swap lines.
So seems like if I run `REMOVE ALL`, I can't `REPLACE` or `KEEP` any tag anymore.

So no idea how can I only concentrate on tags I need with dropping everything that I don't need with one-liner?

Writing documentation for open-source project based on blackbox testing sounds really weird.

vsoch:
ah then the documentation is correct. :)

> Writing documentation for open-source project based on blackbox testing sounds really weird.

I don't understand this comment (it seems like it could be a bit insulting) but I'll just say that if you see an issue in the docs you are welcome to suggest changes. In this case the docs are correct, so there's that.

In practice it's not hugely practical to just wipe every header field from a dicom - many of them are needed. I would suggest you look at your data and come up with reasonable filters (e.g., removing everything with a date, or name, the library supports regular expression matching for that).

Insulting the maintainers of the library for their documentation is not going to get you very far. Just FYI. :)

[...]

PR #198, opened 2021-12-16
WIP 197 remove + keep combo should keep

glebsts:
# Description

`KEEP someField` ignores `REMOVE someField` by adding it to skip list.
`ADD someField` overrides `KEEP someField` as discussed under the issue.
Added tests. Also added test illustrating that `except` is working as regex match (documentation to be updated)
Related issue: #197

# Checklist

- [x] tests added
- [x] documentation updated

### DIFF
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -277,7 +277,9 @@ class DicomParser:
         keeps = []
         if self.recipe.deid is not None:
             keeps = [
-                action.get("field") for action in self.recipe.get_actions(action="KEEP")
+                action.get("field")
+                for action in self.recipe.get_actions(action="KEEP")
+                if action and action.get("field")
             ]
         return keeps
 
@@ -339,7 +341,7 @@ class DicomParser:
 
         Parameters
         ==========
-        field: if provided, a field for expand
+        field: a field for expand
         value: field value
         action: the action from the parsed deid to take
            "field" (eg, PatientID) the header field to process
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -210,11 +210,6 @@ class DicomParser:
         if remove_private:
             self.remove_private()
 
-        # if we loaded a deid recipe, KEEP must overrule REMOVE, adding kept fields to skiplist
-        if self.recipe.deid is not None:
-            for action in self.recipe.get_actions(action="KEEP"):
-                self.skip.append(action.get("field"))
-
         # In the parsing, we generate a list of DicomField objects.
         fields = self.get_fields(expand_sequences=True)
 
@@ -274,6 +269,21 @@ class DicomParser:
             skips = self.config.get("get", {}).get("skip", {})
         return skips
 
+    @property
+    def keep(self):
+        """
+        Return a list of fields to keep, as defined by all KEEP actions in recipe
+        """
+        keeps = []
+        if self.recipe.deid is not None:
+            keeps = list(
+                map(
+                    lambda action: action.get("field"),
+                    self.recipe.get_actions(action="KEEP"),
+                )
+            )
+        return keeps
+
     def get_fields(self, expand_sequences=True):
         """expand all dicom fields into a list, where each entry is
         a DicomField. If we find a sequence, we unwrap it and
@@ -284,7 +294,7 @@ class DicomParser:
                 dicom=self.dicom,
                 expand_sequences=expand_sequences,
                 seen=self.seen,
-                skip=self.skip,
+                skip=self.skip + self.keep,
             )
         return self.fields
 
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -276,12 +276,9 @@ class DicomParser:
         """
         keeps = []
         if self.recipe.deid is not None:
-            keeps = list(
-                map(
-                    lambda action: action.get("field"),
-                    self.recipe.get_actions(action="KEEP"),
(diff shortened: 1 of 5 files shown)


vsoch:
This looks fantastic! It looks like you should run black to fix the formatting issues (black==21.12b0) and then don't forget to increment the version in version.py and add a note to the CHANGELOG.

Do we have except: documented well enough along with the updated use of KEEP?

glebsts:
No, as I mentioned documentation update it still coming, I'd like to agree on logic change per se.  This is reflected in checklist in PR header :)

I looked into failing check, but don't see exact problems there, just that file needs reformatting, so will setup `black` locally and see what it could give me.

vsoch:
okay no worries! And yes you can pip install black (I showed the version used in the CI in the comment above) and then run black deid and it should format. Sorry for the extra trouble - black is a bit fussy sometimes, but it's easier to instill a common format for the code.

deid/dicom/parser.py
```diff
@@ -269,6 +269,21 @@ def skip(self):
...
         return skips
 
+    @property
+    def keep(self):
+        """
+        Return a list of fields to keep, as defined by all KEEP actions in recipe
+        """
+        keeps = []
+        if self.recipe.deid is not None:
+            keeps = list(
```
vsoch, line 279: Generally it's faster to use [] or {} as opposed to the types, and I suspect this is considered Pythonic but it's not easy for the reader to understand. Could you please maintain the same structure you used in the parse function before?

glebsts: Previously I iterated all actions, adding them by one, and it was one-time iteration. Now potentially property can be called multiple times if some new code is added, so using `append` makes no sense (no point to append to already populated). 
I can get rid of lambda, but anyway it should be `keeps = <new array>`, not previously used `append`.
So it would be 
```
keeps =[action.get('field') for action in self.recipe.get_actions(action='KEEP')]
```
As pythonic as it could be and no lambdas :)
Or did you mean smth else?

vsoch: Yep that looks good! I would add a check like "if action and action.get('field')` to make sure the action is defined and it has a field.

glebsts: what could cause it to be undefined? `get_actions` returns empty array even if recipe has no KEEP actions, checked it. But how can it return i.e `[None]`?

glebsts: Nope, it causes other failure in `parser.py:355`, `if re.search("^values", field):`. According to docs `field` is `if provided`, but actually if it is `None` it raises exception.

glebsts: In that sense `pydicom` tries to emulate `dict.get()` in a bad manner - Python's `dict` fails if key not present and `default` not given, but `pydicom` defaults to `default=None`. Meaning `perform_action` will be called with `field == None` if it is missing in recipe, and exception will be thrown.

vsoch: It's a case that I can't forsee but I think it's okay to be extra conservative and make sure that it doesn't happen, e.g., given that someone feeds in a custom listing of actions and has accidentally one without field (or a None).

docs/_docs/user-docs/recipe-headers.md
```diff
@@ -345,6 +345,13 @@ BLANK except:StudyDate|StudyTime
 BLANK except:StudyTime
 ```
 
+_except_ is checking field names against given pattern, so if you for example set
+```
+REMOVE excep:Manufacturer
```
vsoch, line 350:
```suggestion
REMOVE except:Manufacturer
```

vsoch (approved):
Beautiful! Aside from the one small typo in the docs that I added a suggestion for, this should be good to go. The last bit before merge is to:

 - bump the version in deid/version.py 
 - update the CHANGELOG.md with a note for this version

glebsts:
I haven't bump version in my last PR, and this PR goes to 0.2.29-RC. So version could be bumped to that one, yeah, just wonder why it is not done yet :)

vsoch:
oops sorry just forgot about that (it's late and I'm tired!) it looks good to go!

vsoch:
Done and done! https://pypi.org/project/deid/0.2.28/ Thank you @glebsts !

glebsts:
Um.. Is 0.2 28 now rebuilt? I'm confused, what's in public repo, real 0.2.28 or 0.2.29rc1?

vsoch:
Yeah sorry for being confusing! We merged 0.0.28 into master but didn't release on pypi because I was waiting for this fix under that version. So the current version in the repo is 0.0.28.

vsoch:
ah crap this was merged into your other PR! I totally missed that. Yes we are going to need to do another release.

vsoch:
So it's up to you how you want to proceed. If you want to do other changes to that branch then go ahead. If you are ready for that branch to PR to master here, then go ahead. I was planning on waiting to release 0.0.28 until we had these changes, but now that will be 0.0.29. Sorry for the confusing-ness, our conversation has happened slowly over time and honestly I just forgot about the plan because I maintain a lot of projects.

glebsts:
No worries! Thank you :) 
Yeah, I followed same branch as we agreed last time with pydicom bump, so..
Have some rest! Merry Christmas, and see you next time :)

vsoch:
Ok, so I’ll do a PR with this branch so we can get it into the mainline, probably later tomorrow. Thank you and merry Christmas! 🎁🎄

PR #208, opened 2022-04-14
Adding ability for deid to support deid-provided functions

### LINKED ISSUE #207: Add deid provided functions
Discussion here: https://github.com/pydicom/deid/pull/203#issuecomment-1071870150

@moloney did you say you wanted to tackle this, or would you like someone else to take a first shot?

vsoch:
As of this version, a user can specify a value as a deid_func:<name> meaning that we use a deid provided function. This is useful for providing some general uid and a customized jitter function, along with other functions that users might want to add. With this change I have provided docs and a contributing section for how to do this, along with running pyflakes on
the tests.

This will close #207 and will assist with #206 

### DIFF
--- a/docs/_docs/contributing/code.md
+++ b/docs/_docs/contributing/code.md
@@ -16,7 +16,7 @@ See the repository `CONTRIBUTING.md` for these same details.
 ## Contributing a Custom Function
 
 Deid ships (as of version 0.2.3) with deid-provided functions that can be used in
-header parsing. To contribute a recipe you should do the following:
+header parsing. To contribute a custom function you should do the following:
 
 
 1. Add a function to deid/dicom/actions, ideally in the appropriate file (e.g., uid functions in uuid.py, etc)
--- a/docs/_docs/contributing/code.md
+++ b/docs/_docs/contributing/code.md
@@ -12,3 +12,55 @@ Generally, for code contribution you should:
 4. Open a pull request against the master branch.
 
 See the repository `CONTRIBUTING.md` for these same details.
+
+## Contributing a Custom Function
+
+Deid ships (as of version 0.2.3) with deid-provided functions that can be used in
+header parsing. To contribute a recipe you should do the following:
+
+
+1. Add a function to deid/dicom/actions, ideally in the appropriate file (e.g., uid functions in uuid.py, etc)
+2. Ensure your function is added to the lookup in `deid/dicom/actions/__init__.py` so it can be found.
+3. Add a test to `deid/tests/test_dicom_funcs.py` that ensures your function works, with or without custom variables.
+
+
+Generally, a custom function should accept the following variables:
+
+ - dicom: the dicom file
+ - item: expected to be the dictionary lookup of user provided values
+ - field: the dicom field
+ - value: the value to replace
+ 
+You can generally define a catch all `**kwargs` if you don't need a field. Finally,
+if you do provide a custom variable, you'll need to also provide a default (or exit on error
+if it's absolutely essential). As an example, if your custom function in the lookup is named
+`generate_sesame_street_character` the user might provide a custom argument as follows:
+
+```
+%header
+
+REPLACE fields:PatientID deid_func:generate_sesame_street_character name=elmo
+```
+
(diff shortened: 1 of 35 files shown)


vsoch:
Ping @moloney and @wetzelj !

moloney:
Sorry for dropping off on this, I am taking a look now.  One immediate comment:  why not use the pydicom "generate_uid" function as I had done previously?  It can replace all the code in "deid/dicom/actions/uids.py", and it can provide purely random UIDs as you are doing in your code or it can provide a consistent mapping (same input UID produces same anonymized UID) so that study / series UIDs remain consistent after anonymization without any coordination between the processing of the separate files. This latter functionality is critical for many use cases.

Thanks for the hard work here!

vsoch:
@moloney that's a fantastic idea! I 100% agree, and I feel silly that I didn't know about it. Can you suggest a name / args etc of how it should look?

moloney:
There are just two optional args, a prefix (defaults to the pydicom root prefix) and an optional list of things to use as entropy when generating the UID (if nothing is provided it just uses system randomness).  If you pass the original UID in as the only entropy source you will get a stable result across all files (same input results in same output).

Docs are here: https://pydicom.github.io/pydicom/dev/reference/generated/pydicom.uid.generate_uid.html

So I guess the deid func should just take two optional args, a prefix and a toggle to enable / disable stable remapping (i.e. passing in the original UID) which I think should default to being turned on.

vsoch:
@moloney for the remapping bit - does that mean passing the original value as the prefix or in the list of entopy items (it sounds like the second for the stability?) I just want to double check before writing it!

vsoch:
okay I think I have it figured out! I'll have an update to the PR for you after work today.

vsoch:
@moloney all set! https://github.com/pydicom/deid/pull/208/commits/bca9596ea2766bac7cb9838a9b04f265156d3d5e

docs/_docs/contributing/code.md
```diff
@@ -12,3 +12,55 @@ Generally, for code contribution you should:
 4. Open a pull request against the master branch.
 
 See the repository `CONTRIBUTING.md` for these same details.
+
+## Contributing a Custom Function
+
+Deid ships (as of version 0.2.3) with deid-provided functions that can be used in
+header parsing. To contribute a recipe you should do the following:
```
wetzelj, line 19: Trivial comment - but this should probably read: "To contribute a **custom function**..."

@vsoch - Sorry for my delay looking at this and responding. I've continued to be pulled in other areas so haven't had much time to spend on deid...

vsoch: Totally no worries! I'll be around when you are ready :)

docs/_docs/examples/func-replace.md
```diff
@@ -10,6 +10,8 @@ change field with a value derived from a function. This example
 was derived based on a prompt in [this pull request](https://github.com/pydicom/contrib-pydicom/pull/14).
 If you are interested in the code for this example, it's available
 [here](https://github.com/pydicom/deid/tree/master/examples/dicom/header-manipulation).
+If you are interested in the functions provided by deid (and you don't want to write your
+own function) see [this documentation](https://pydicom.github.io/deid/user-docs/recipe-groups/).
```
wetzelj, line 14: The link on "this documentation" is currently referencing recipe-groups, this should be a link to recipe-funcs.

wetzelj (commented):
I've looked through everything in this PR and commented on a couple trivial things. I won't be able to pull this in do my app to do any testing for a bit, but honestly, I think these changes are great!  I like the idea of having DEID provided custom functions - and we'll have a few to contribute once we have an opportunity to clean them up for submission.

docs/_docs/examples/func-replace.md
vsoch: That's called me failing at copy paste :laughing:

vsoch: Fixed.

vsoch:
heads up y'all! I'm going to merge this end of day - I think it looks fairly good and is an improvement to deid and we've had overall positive review.

wetzelj:
Sounds good to me!

PR #212, opened 2022-09-09
Add group:xxxx field expander

jstorrs:
# Description

This is an attempt to add a group: field expander to support some CTP-style rules that target tag groups. The idea is to support rules such as:

`KEEP group:0018
`
# Checklist

# Open questions

~~I've tested this interactively, but unfortunately, I haven't yet figured out how to get pytest to run on deid. It keeps complaining that deid.tests is an unknown module (see attached). I'm new to pytest so I'm probably messing something obvious up.~~

~~[pytest-errors.txt](https://github.com/pydicom/deid/files/9539088/pytest-errors.txt)~~


jstorrs:
Oh, it's not pytest... do you run them:

`python -m unittest deid/tests/test*.py`

vsoch:
I always check the testing file to remember the command!

https://github.com/pydicom/deid/blob/6dd9c7e19a61c46e231c827163de74815628697a/.github/workflows/main.yml#L39-L43

vsoch:
In what cases would someone have an identifier and not the name?

jstorrs:
The various DICOM tag are organized into functional groups.

In CTP for example one of the rules that are available is to keep an entire group. See for example the "k" element in CTP scripts:

https://github.com/johnperry/CTP/blob/master/source/files/examples/example-ctp-dicom-anonymizer.script
https://github.com/johnperry/CTP/blob/master/source/files/examples/example-tfs-dicom-anonymizer.script

```
<k en="T" t="0018">Keep group 0018</k>
<k en="T" t="0020">Keep group 0020</k>
<k en="T" t="0028">Keep group 0028</k>
```

In CTP, I can't remember if the documentation discusses the script files directly because in CTP you're intended to use the configuration utility and not the XML directly. But the k elements work on entire groups specified by the t attribute. en is whether that particular rule is enabled. Groups 18, 20 and 28 are often useful for this sort of targeting.

https://mircwiki.rsna.org/index.php?title=The_CTP_DICOM_Anonymizer

vsoch:
Oh neat!

So the formatting tests (that are failing) are fairly simple - you run black (new version) on deid.

```
black deid
```
and that should fix the test failure. And then just add a note to the CHANGELOG.md and bump the version in deid/version.py and we will be good to go!

vsoch:
This looks great! :partying_face: 

I think we are just about ready for merge. The last little bit (which I forgot in my last comments, sorry about that) is to add a few sentences and a quick example to the user facing docs - I think maybe on this page? https://pydicom.github.io/deid/user-docs/recipe-headers/

docs/_docs/user-docs/recipe-headers.md
```diff
@@ -324,6 +324,19 @@ BLANK contains:Name
 Notice how we get Name in uppercase (when our search string was lowercase) and
 it can appear anywhere in the field.
 
+**group**
+
+The group filter selects fields based on their DICOM group. Groups are
+specified as hexidecimal numbers (of up to four digits). For example,
+the following rulea will keep all elements contained within DICOM
```
vsoch, line 331:
```suggestion
the following rules will keep all elements contained within DICOM
```

docs/_docs/user-docs/recipe-headers.md
```diff
@@ -324,6 +324,19 @@ BLANK contains:Name
 Notice how we get Name in uppercase (when our search string was lowercase) and
 it can appear anywhere in the field.
 
+**group**
+
+The group filter selects fields based on their DICOM group. Groups are
+specified as hexidecimal numbers (of up to four digits). For example,
```
vsoch, line 330: Typo caught by the CI! 

```suggestion
specified as hexadecimal numbers (of up to four digits). For example,
```

vsoch (approved):

vsoch:
Thanks @jstorrs ! This is a cool little addition (very useful!) and I'll get it released soon.

PR #215, opened 2022-09-11
Add vr:XX field expander

jstorrs:
# Description

This is adds a vr: field expander that allows fields to be selected by VR. This allows targeting based on content. For example vr:PN will select all Person Name elements. Or vr:TM will select all Time elements without selecting Date Time elements.

Note: endswith:Time will select some fields that are DT (such as AcquisitionDateTime)

# Checklist

# Open questions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
- - add vr:XX field expander to select elements by VR (0.2.32)
+ - add vr:XX field expander to select elements by VR (0.2.33)
  - add group:XXXX field expander to select all elements with a specified DICOM tag group (0.2.32)
  - custom class example for using dicom.Dataset, not requiring on client init [#211](https://github.com/pydicom/deid/pull/211) (0.2.31)
  - adding support for deid provided functions [#207](https://github.com/pydicom/deid/issues/207) (0.2.3)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - add vr:XX field expander to select elements by VR (0.2.32)
  - add group:XXXX field expander to select all elements with a specified DICOM tag group (0.2.32)
  - custom class example for using dicom.Dataset, not requiring on client init [#211](https://github.com/pydicom/deid/pull/211) (0.2.31)
  - adding support for deid provided functions [#207](https://github.com/pydicom/deid/issues/207) (0.2.3)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
- - add vr:XX field expander to select elements by VR (0.2.33)
+ - add select:vr:XX field expander to select elements by VR (0.2.33)
+ - rename group:XXXX field expander to select:group:XXXX
  - add group:XXXX field expander to select all elements with a specified DICOM tag group (0.2.32)
  - custom class example for using dicom.Dataset, not requiring on client init [#211](https://github.com/pydicom/deid/pull/211) (0.2.31)
  - adding support for deid provided functions [#207](https://github.com/pydicom/deid/issues/207) (0.2.3)
--- a/deid/dicom/fields.py
+++ b/deid/dicom/fields.py
@@ -86,17 +86,22 @@ class DicomField:
                 return True
         return False
 
-    def tag_within_group(self, group):
+    def select_matches(self, expression):
         """
-        Determine whether the element tag belongs to a group
+        Determine whether the element has a specific selected attribute
         """
-        return self.element.tag.group == group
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+ - add vr:XX field expander to select elements by VR (0.2.32)
```
vsoch, line 17:
```suggestion
 - add vr:XX field expander to select elements by VR (0.2.33)
```

docs/_docs/user-docs/recipe-headers.md
```diff
@@ -337,6 +337,16 @@ KEEP group:0020
 KEEP group:0028
 ```
 
+**vr**
```
vsoch, line 340: Interesting - so @jstorrs I'm wondering, how about for direct attributes doing a more generic select match? E.g., instead of hard coding each attribute (there could be many) to match the overall design of the library, why not have it under a `select` namespace? For example, for vr:

```
BLANK select:vr:TM
```
And then the code is flexible to accept and compare any sets of direct (or even nested) attributes, instead of needing to hard code as we do now. E.g.,

```python

    def select_matches(self, expression):
        """
        Determine whether the element has a specific selected attribute
        """
        attribute, value = expression.split(':', 1)    
        if hasattr(self, attribute.upper(), False):
            return getattr(self, attribute.upper()) == value
        return False
```
And if we wanted to handle nesting, we'd do it based on levels of `:` (e.g., spit into a list and keep going until one is left, and that is the value, and that assumes the value does not have a `:` itself. It might make sense to have a nice clean little class to handle holding the raw value and then parsing it and retrieving the final value. We would instantiate it when we find something that has the "select" prefix.

Does that make sense? Let me know what you think. If we want to include group under this schema we could have an additional lookup of the nesting to match to an identifier the user provides. 

The high level idea is that if we can do something flexible to any attribute someone might be interested in we can avoid a hard coded function for each one.

jstorrs: I'll need to think a bit about this (sorry I can be slow). The sort of basic features I would expect to access are what would be seen on the output of say DCMTK's dcmdump. Those will be the group/element numbers, the VR and the VM. But I've personally never encountered needing to access VM in an anonymizer script. Group and VR and pretty useful on their own. VR is nice for tying into lookup tables and databases. Group is helpful to abtract out some modality features that are almost never a PHI concern. So VR is actually the last one I have in mind.

But beyond those my preference actually would be to allow user-provided functions to be registered. They would just look at the pydicom element and return true/false if they wanted to apply. All sorts of complex logic could be put in these sorts of functions. But sequences need some thought. You'd probably also want user-functions to be able to suggest whether they want to descend into sequences or not.

vsoch: okay sounds good - ping me when you want to chat more, or if you want to explicitly layout what you want (and I can try for a first shot).

And no need to apologize - there is lots to do and I totally get it!

jstorrs: Ok, I think I see what you are suggesting now and it seems reasonable to me. The rules certainly are more self-explanatory with the select: namespace. I can at least migrate to the namespaces such as select:group:XXXX and select:vr:XX while we think about implementation.

vsoch: And if we think the `<attribute>:<value>` is more elegant, we could have similar, without the select group, and if a `:` (or maybe double) is detected it would be flagged as a select to check. Just an idea! I really like how elegant your original `group:<value>` and `vr:<value>` are and don't want to lose sight of that.

jstorrs: I think separating out the select into its own function makes the code more compact and easy to follow.

vsoch: Agree! And then the next question is what the user interaction (what they write in the file) looks like to get there.

jstorrs: I think writing "select:property:value" in the file is easy to remember and to follow. Once people know the select:property:value pattern they'll at least know that "group" and "VR" are properties.

Potentially the value component might be extended to handle multiple values in the future (I may have uses for that, but I'll see what I think when I get to that step). Things like ranges "select:group:6000-60FF" to select overlay planes seem likely. IIRC CTP's approach was a special rule to remove overlays but targeting them could be interesting. Or maybe sets to select multiple VR with something like "select:VR:DT|TM|DA").

Maybe this could might even be extended to select:property1:value1:property2:value2. But I'm not sure I have an immediate use for that and I don't want to add things just for the sake of adding things. I'll keep it in the back of my mind in case something comes up where I'm wishing I could apply it.

vsoch: I think what might be a good idea is to step away for a few days, and in the time write down the explicit examples that you really want to work - the use cases that are strongest for you. E.g., likely they are the first set of things, and the expanded one might be developing for something that isn't needed.

deid/dicom/fields.py
```diff
@@ -86,11 +86,22 @@ def value_contains(self, expression):
...
         return False
 
-    def tag_within_group(self, group):
+    def select_matches(self, expression):
         """
-        Determine whether the element tag belongs to a group
+        Determine whether the element has a specific selected attribute
         """
-        return self.element.tag.group == group
+        attribute, value = expression.split(":", 1)
```
vsoch, line 93: Yes! This is exactly what I had in mind - it matches the pattern of the other selectors because it's a generic term, e.g., func, startswith, etc.

vsoch:
heyo! Did you have any more thoughts on this design? I think it's good to go in for a first shot - we had a request in #219 for the feature so it would be good to have!

jstorrs:
I agree the latest "select:vr" version is good to go.

vsoch (approved):

Issue #225, opened 2022-09-28
Really heavy package


RobinFrcd:
Hello,
I realized that the package is quite big (**27.2 MB** for the .tar.gz), would it be possible to lighten it ?

I assume the culprit here are the `.dcm` in the [data folder](https://github.com/pydicom/deid/tree/master/deid/data). This folder is only used for tests, right ? Would it be possible to move the tests and tests data out of the package folder in order not to ship it during the build ?

Thanks :pray:

vsoch:
Agree! There is already an issue for this, but nobody has worked on it yet. https://github.com/pydicom/deid/issues/167 Ping @wetzelj to see if there is still interest. Closing here and we can continue conversation in https://github.com/pydicom/deid/issues/167

Issue #167, opened 2020-12-07
Refactor Unit Tests to utilize separate data store for sample images


wetzelj:
Created this issue to continue discussion started on PR #166:

> > With this proposed new issue, I believe that I have an opportunity to build out our test image data set with a larger variety of PHI-safe test images from a variety of makes/models of scanners. Our department has an annual employee art show in which we can image (XR, CT, US) objects to serve as the basis for works of art.
> 
> I really like this idea. And actually, we can fairly easy separate the small example data (e.g., dicom cookies) from those that are just used for testing, which arguably shouldn't even be stored in the repository because they are so large. I think what would make sense to do is follow the convention of [pydicom-data](https://github.com/pydicom/pydicom-data), and even first ask if there is room to accept files for deid here to that library (so pydicom can have a single, shared data library). @darcymason would this be something we could consider? If we want to keep the libraries separate, and you'd want to make a new dataset not just for deid but for other dicom enthusiasts then we could also have it branded differently and then used by the library here. Either way, we could definitely have a function to decompress after install.

Having looked closer at [pydicom-data](https://github.com/pydicom/pydicom-data), I wonder if it would be better for us to first attempt to convert the unit tests to use images that are already stored within pydicom-data?  While I haven't done an exhaustive look to determine if all of our test cases could be covered by images in pydicom-data, it does appear that there's high potential that they could be.  This could eliminate the need for a standalone deid test image repo.

vsoch:
@wetzelj I think this is a great idea - and we could still keep (the smaller) files alongside deid in case anyone wants a quick dataset to play with. If there is something missing, I'm hoping that @darcymason would be open to a PR for us to add it. That way, both repos can share the same testing data.

wetzelj:
What are your thoughts on potentially upgrading pydicom beyond the current required version 1.3.0 with this? 

For about two weeks now, I've been running the latest deid code with pydicom 2.1.1 and have seen no issues with this upgrade.  Between 1.3.0 and 2.1.1 there have been some changes to [pydicom.data](https://pydicom.github.io/pydicom/stable/reference/data.html) which could impact how this issue is implemented.

vsoch:
[...]

PR #226, opened 2022-09-28
Move data and tests out of the package

RobinFrcd:
# Description

Related issues: #167 #225 

The idea here is to move the heavy data folder out of the package.
Moving the tests folder outside the package also makes sense because it's not needed to run any function.


vsoch:
oh hmm... this could work too. What are the benefits of this approach vs. having the data as a separate (installable) package? indeed this one is probably easier to maintain! 

And you are so speedy! I'm barely woken up :)

deid/data/__init__.py
```diff
@@ -11,32 +11,4 @@
 
 import os
 
-from deid.logger import bot
-from deid.utils import get_installdir
-
 data_base = os.path.dirname(os.path.abspath(__file__))
```
vsoch, line 14: Is this data_base used anywhere? E.g., why are we keeping it?

tests/Xtest_dicom_header.py
```diff
@@ -13,8 +14,8 @@
 import tempfile
 import unittest
 
-from deid.data import get_dataset
 from deid.utils import get_installdir
+from tests.common import get_dataset
```
vsoch, line 18: This should ideally be relative so we aren't counting on the pythonpath or present working directory have a folder "tests"

tests/test_clean.py
```diff
@@ -30,7 +29,7 @@
 class TestClean(unittest.TestCase):
     def setUp(self):
         self.pwd = get_installdir()
-        self.deidpath = os.path.abspath("%s/tests/resources/" % self.pwd)
+        self.deidpath = os.path.abspath("%s/../tests/resources/" % self.pwd)
```
vsoch, line 32: This should be relative too - use the path of the file plus the path to resources instead of self.pwd.

vsoch (commented):
This isn't what I originally had in mind, but I think it's a great idea!

I do think the imports should generally be relative. And the larger thing we need to fix is all the examples in the docs that show how to get a dataset and use it - if we aren't installing the data to the library (which would have been my idea and maintaining it's location) we can't show them that.

vsoch:
And I'll share my idea - in case you think it makes sense! I was going to make an installable package, and then if the tests are attempting to be run or any import of testdata is attempted and the path doesn't exist yet, we clearly show the user "You need to pip install deid-data" (or similar). Then all the testing logic and paths would stay the same, but the library would not ship by default with the data (but it could be installed, if desired).

deid/data/__init__.py
RobinFrcd: I think it's still used to get `deid.dicom` for example, right ?

tests/Xtest_dicom_header.py
RobinFrcd: I've replaced with `tests_commons` is it ok for you ? Seems like relative imports + unittest don't work that well together :hot_face:

deid/data/__init__.py
vsoch: Right - I was thinking those deid.dicom would stay with the library, and just the larger datasets would be externally installed.

RobinFrcd:
> And I'll share my idea - in case you think it makes sense! I was going to make an installable package, and then if the tests are attempting to be run or any import of testdata is attempted and the path doesn't exist yet, we clearly show the user "You need to pip install deid-data" (or similar). Then all the testing logic and paths would stay the same, but the library would not ship by default with the data (but it could be installed, if desired).

I guess it makes sense, sure ! I also saw in the comment "In the future, we can add https endpoints", it may also be a legit solution !

deid/data/__init__.py
RobinFrcd: Yes, it totally makes sense to keep this file!

vsoch:
@RobinFrcd take a look at my idea here: https://github.com/pydicom/deid/pull/227

It's a bit simpler in the sense that deid-data is just an external package that anyone can install, and deid uses it for tests (and someone could arguably use it for their own use case). I think it might be preferable to have this format, primarily so someone can easily install the data on a whim without cloning, and we can cleanly maintain the two projects (https://github.com/pydicom/deid-data). Let me know your thoughts!

RobinFrcd:
@vsoch Sure, it makes sense. If you don't mind maintaining 2 packages ! Looks good to me :+1:

vsoch:
@RobinFrcd I have hundreds of projects and like 90 something just on pypi? That's kind of what I do :) :laughing:

vsoch:
@RobinFrcd I'm deciding to go with the other implementation (a long time contributor also preferred it) but I want to say how grateful I am that you jumped on this! And it was all before I woke up! So thank you hugely and kindly - we won't get to use it but it was a really good idea.

RobinFrcd:
Sure, no problem ! I'm glad we went from 20MB+ to less than 50kB !! :rocket:

vsoch:
I know amazing! Lol I feel so badly... should have done this sooner :laughing:

Issue #230, opened 2022-10-01
Specify more than one coordinates in the filter


NirutaDhimal:
I am trying to remove clean pixels of 2 -3 coordinates of a dicom file. How do I specify it in the filter ?
[
![err2](https://user-images.githubusercontent.com/58328226/193401096-f88c1d37-c1c2-401d-9ab2-d35d4245aa3b.png)
This is my filter But I want to clean pixels on bounding box 4,462,23,102. How do I add this in filter?

vsoch:
Here is an example of an entry with two coordinates: https://github.com/pydicom/deid/blob/master/deid/data/deid.dicom#L301-L308

NirutaDhimal:
I tried that but It's not working. Only the First coordinates were removed.

vsoch:
That's a bug then! Are you able to share some dummy data so I can use for a debuggig/test case?

vsoch:
Actually I have a good test case already! I should be able to work on this and propose a fix today - stay tuned!

NirutaDhimal:
I don't have any dummy data. Sorry for that. Thanks

vsoch:
@NirutaDhimal I tested the syntax with multiple coordinates and it appears to work as I'd expect (both sets show up):

```
In [1]: results['results']
Out[1]: 
[{'reason': 'and Manufacturer contains Philips Modality contains US ImageType contains Cardiology',
  'group': 'graylist',
  'coordinates': [[0, '0,0,1024,23'], [0, '0,0,1024,23']]}] # notice there are two here
```
I think very likely the issue is that you either have a bug in the bounding box you are asking for, or a "contains" statement that doesn't apply to your image. It's hard to debug beyond that without seeing your data. However, the format for the coordinates that I showed above is correct.

NirutaDhimal:
@vsoch I printed client.detect() result and both set shows up in the coordinates. But when I did client.save(), the saved image still contains the text in the 2nd coordinate.
[
![err3](https://user-images.githubusercontent.com/58328226/194198629-7e0ecccd-dbe9-4a50-afb5-380f1886874c.png)
](url)

vsoch:
Did you do `client.clean()` ?

NirutaDhimal:
Yes, I did that. First I did client.detect() and then client.clean() and at last I did client.save()

vsoch:
[...]

PR #234, opened 2022-10-13
add support for translating ctp coordinates

vsoch:
We identified a bug in #230 where the ctp coordinates are supplied as:

```
(xmin, ymin, width, height)
```
but our parser is expecting:
```
(xmin, ymin, xmax, ymax)
```
So I think the translation is fairly simple - the first two coordinates (I think) are correct, and we just need to add the width and height to each to derive the xmax and ymax that we expect. We can do this right at parsing (config/utils.py) and then still provide the "same" structured coordinates later.

This means I've added a set of new coordinate types, `ctpcoordinates` and `ctpkeepcoordinates` that mirror the `coordinates` and `keepcoordinates` but expect this slightly different input format, and updated our default deid.dicom to use them (since I believe all entries in here minus the ones from stanford faculty are from CTP!)

# Checklist

# Open questions

Is there a reasonable way to better test this?


deid/config/utils.py
```diff
@@ -258,9 +258,40 @@ def parse_filter_group(spec):
...
+        return raw
+    new_coordinate = [int(x) for x in raw.split(",") if x]
+
+    # Cut out early for malformed coordinate
+    if len(new_coordinate) != 4:
+        bot.exit("Coordinates are expected to have length of 4, found %s" % raw)
+    xmin, ymin, width, height = new_coordinate
+
+    # Translate CTP coordinate to the convention we use
+    return [xmin, ymin, xmin + width, ymin + height]
```
wetzelj, line 289: There was a little bug with this return.  Downstream locations expect coordinates to be a string comma separated list, not an actual list of ints. I've commited a fix in with PR #237.

vsoch:
@wetzelj good to go on your end?

wetzelj:
I've run a couple basic tests - and this looks pretty good. 

At first I was a little concerned about the REMOVE functionality, specifically [this comment](https://github.com/pydicom/deid/pull/204#issuecomment-1072435662) from the discussion on #197.  When I tested placing a REMOVE at the end of my recipe, it did indeed remove a field that I had already added.  The REMOVE wasn't as I expected from reading the comments on #197.  On some further testing, it appears as though a REMOVE following an ADD or JITTER is still removing the field (what I consider current functionality), while a REMOVE following a KEEP or REPLACE ignores the remove (new functionality, but very worthwhile). 

That said, I don't know that this necessarily needs to hold up this merge, but should probably be opened as another issue - to thoroughly evaluate and determine how we want these all to work together.    What do you think?

vsoch:
I think I would expect a `REMOVE` to remove (regardless of previous history) UNLESS someone asked to keep - so I don't see issue with the current functionality.

vsoch:
I think there are enough good changes here that it's worth merging and releasing, and we can open an issue for further discussion on remove. Thanks @wetzelj !

wetzelj:
Thank you @vsoch!

PR #199, opened 2021-12-20
0.2.29 rc

vsoch:
# Description

@glebsts just opening this so when you have tested and are ready we can merge into master. Definitely no rush, we can do after the holiday! I just want to make sure I don't forget :)

PR #237, opened 2022-11-21
Add/remove keep behavior (Previous PR #199 - 0.2.29 rc)

wetzelj:
# Description

Related issues: #197, #198, #204, 

Whitelisting was not possible for replacing. Now fields being replaced or jittered are excluded from removal (either REMOVE ALL or REMOVE SomeField).

It is done using separate property and private variable to keep result (lazy eval).

Made clarification regarding KEEP list - those fields are being removed from field list to process, meaning that KEEP SomeField + REPLACE SomeField ... won't work, as kept field is being excluded from list of fields to process for other actions.

Added tests.
Bumped release candidate number and changelog (+small fix for previous change)

Minor docs update.

# Checklist

# Open questions

### DIFF
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -32,6 +32,11 @@ class DicomParser:
     file. For each, we store the element and child elements
     """
 
+    """
+    list of fields to exclude from remove because they are replaced and jittered
+    """
+    _excluded_fields = None
+
     def __init__(
         self, dicom_file, recipe=None, config=None, force=True, disable_skip=False
     ):
@@ -280,6 +285,41 @@ class DicomParser:
             skips = self.config.get("get", {}).get("skip", {})
         return skips
 
+    @property
+    def keep(self):
+        """
+        Return a list of fields to keep original, as defined by all KEEP actions in recipe
+        Those fields are not impacted by REPLACE/JITTER actions
+        """
+        keeps = []
+        if self.recipe.deid is not None:
+            keeps = [
+                action.get("field")
+                for action in self.recipe.get_actions(action="KEEP")
+                if action and action.get("field")
+            ]
+        return keeps
+
+    @property
+    def excluded_from_deletion(self):
+        """
+        Return once-evaluated list of fields that are not removed by REMOVE ALL or REMOVE SomeField,
+        as they later have to be changed by REPLACE / JITTER
+        That allows whitelisting fields from REMOVE ALL/SomeField to change them if needed (i.e. obfuscation)
+        """
+        if self._excluded_fields is None:
+            self._excluded_fields = []
+            if self.recipe.deid is not None:
+                self._excluded_fields = [
+                    action.get("field")
+                    for action in (
+                        self.recipe.get_actions(action="JITTER")
+                        + self.recipe.get_actions(action="REPLACE")
+                    )
+                    if action and action.get("field")
+                ]
+        return self._excluded_fields
+
     def get_fields(self, expand_sequences=True):
         """expand all dicom fields into a list, where each entry is
         a DicomField. If we find a sequence, we unwrap it and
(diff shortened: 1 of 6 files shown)


deid/config/utils.py
```diff
@@ -284,9 +284,11 @@ def derive_ctp_coordinate(raw):
     if len(new_coordinate) != 4:
         bot.exit("Coordinates are expected to have length of 4, found %s" % raw)
     xmin, ymin, width, height = new_coordinate
+    new_coordinate[2] = xmin + width
+    new_coordinate[3] = ymin + height
 
     # Translate CTP coordinate to the convention we use
-    return [xmin, ymin, xmin + width, ymin + height]
+    return ",".join([str(i) for i in new_coordinate])
```
vsoch, line 291: Do we want to return a string here? Coordinates should technically be numbers. I think it would make sense to convert to string only when we need that, or perhaps have the default return as numbers, and have an argument that asks for a string format.

deid/dicom/parser.py
```diff
@@ -280,6 +285,41 @@ def skip(self):
...
+        return keeps
+
+    @property
+    def excluded_from_deletion(self):
+        """
+        Return once-evaluated list of fields that are not removed by REMOVE ALL or REMOVE SomeField,
+        as they later have to be changed by REPLACE / JITTER
+        That allows whitelisting fields from REMOVE ALL/SomeField to change them if needed (i.e. obfuscation)
+        """
+        if self._excluded_fields is None:
```
vsoch, line 310: For this I usually prefer a pattern like:

```python
self._excluded_fields = self._excluded_fields or []
```
instead of the nesting.

deid/dicom/parser.py
```diff
@@ -518,7 +558,7 @@ def _run_action(self, field, action, value=None):
                     funcs=self.deid_funcs,
                 )
 
-            if do_removal is True:
+            if do_removal is True and not (field.name in self.excluded_from_deletion):
```
vsoch, line 561:
```suggestion
            if do_removal is True and field.name not in self.excluded_from_deletion:
```

vsoch (commented):
Just a few tweaks! And @wetzelj your formatting reminds me that we should add pre-commit here - I've been using it (with dev environments for VS Code optionally) for other Python projects and it's :kissing_heart: chef's kiss! When we finish up this work I can put this on my queue to do - it makes formatting / linting much easier than what we are doing now.

deid/config/utils.py
wetzelj: It depends. :) 

Returned as a list of ints, this new function does not work with the existing deid code for coordinates.  When I started testing with this, specifying ctpcoordinates ended up with an exception ('int' object has no attribute 'lower') because the output of this function is eventually passed into [clean_pixel_data](https://github.com/pydicom/deid/blob/27d46bb8851952fc3c6639d4675e254907085af7/deid/dicom/pixels/clean.py#L329) which performs a .lower() on it - expecting that it's a string.

When I received this exception, I targeted the new code for a fix - rather than looking to refactor all of the other places where the code currently generates this string.

vsoch: Gotcha, and it looks like it's only used within that utils file:

```
$ grep -R derive_ctp_coordinate
Binary file deid/config/__pycache__/utils.cpython-39.pyc matches
deid/config/utils.py:def derive_ctp_coordinate(raw):
deid/config/utils.py:            coordinate = derive_ctp_coordinate(
deid/config/utils.py:            coordinate = derive_ctp_coordinate(
```
How about let's make it a private function (e.g., just rename to have a single underscore) and call it a day. We mostly don't want a user finding it and using it.

wetzelj: Done!

vsoch:
Thanks @wetzelj !

Issue #241, opened 2022-11-28
BLANK Action only applicable to US and SS VRs


wetzelj:
While performing some testing, I noticed that the BLANK action is currently limited to only the SS (Signed Short) and US (Unsigned Short) VRs.  Do you think that this restriction can/should be removed?

vsoch:
If the others are BLANK-able, then yes it should! I don't remember why this would be the case - it must have been added and I missed the error. Feel free to open a PR to fix!

wetzelj:
Ok... Feel free to assign this to me.  I've already started working on it along with a set of new unit tests that I've been thinking about.  

I want to have a very explicit set of tests which validates the interactions between two rules on the same field - and then I'll update the docs with a table of rule interactions and expected results.

PR #242, opened 2022-11-30
Bug Fixes and Unit Tests for BLANK & REPLACE Actions & Sequences

wetzelj:
# Description
1. Updated BLANK action to function on all VRs.
2. Added unit tests to ensure BLANK functions on all newly added VRs.
3. Added unit tests to validate multiple actions on a single field.  Added documentation which describes the current interactions. 
4. Corrected issue with REPLACE on numeric VRs (and URI as well).
5. Added unit tests to validate REPLACE on all (applicable?) VRs.  Currently SQ, and the O* VRs are excluded. 
6. Corrected issue preventing actions from executing on fields within sequences. 

Related issues: 

- #241 
- #244
- #243 

# Checklist

# Open questions
1. For the interaction unit tests and documentation, I simply test and document what the library currently does as of v0.3.1.  I would appreciate a closer review of this.  As I was building these tests, some of the interactions were not what I expected... but that may just be my mind working a little differently than others! :)


wetzelj:
The testing is failing due to the changes that were merged into deid-data.  The CI testing is still using the old image.

vsoch:
Old image == old release?

vsoch:
@wetzelj if you can do a PR to bump the version / changelog over there, I can make a new release!

wetzelj:
Yeah... sorry about that.  I just realized I didn't bump the version of deid-data. Will do it now.

docs/_docs/user-docs/recipe-headers.md
```diff
@@ -251,6 +251,80 @@ The above would remove everything except for the pixel data, and a few fields
 that are relevant to its dimensions. It would add a field to indicate the
 patient's identity was removed.
 
+The table below shows the full details of how multiple actions defined on the same
```
wetzelj, line 254: @vsoch - As an additional check, I ran the action interaction tests against 0.2.28.  While I expected the tests that involved KEEP and BLANK to fail, there were a few others which had altered behavior.  The image below is the 0.3.1 action interaction grid added to the documentation.

- The cells in green indicate those tests which functioned the same in 0.2.28 vs 0.3.1.
- The cells in yellow indicate the tests which run successfully in 0.3.1, but **expectedly** fail in 0.2.28 due to KEEP/BLANK issues.
- The cells in red are those that run successfully in 0.3.1, but fail in 0.2.28.

image

Given the "precedence" defined, I think that the cells in red are actually okay too, since per the precedence of the actions, JITTER and REPLACE are both > REMOVE.   While this interaction has changed since 0.2.28, I believe this is a correction in behavior. 

I'd just like to make sure that I'm thinking correctly on all of this and we don't have a bug that needs to be addressed.

vsoch:
okay sorry for delay - went into meetings and just coming out! I just merged/released deid-data and did a re-run of the test.

docs/_docs/user-docs/recipe-headers.md
vsoch: This is so pretty! So the table is saying when you do a "thing in row index" then a "thing in column index" there is some underlying test that is successful / failed / no change. So for the red ones that means:

- Jitter then remove has changed functionality
- Replace then remove has changed functionality
- Remove then jitter and remove then replace is changed.

The last row makes sense - if Jitter and replace are preferred over remove, then we shouldn't see them removed. And then the opposite pattern for the other two. So yes I think this makes sense.

deid/dicom/parser.py
```diff
@@ -167,7 +167,7 @@ def get_nested_field(self, field, return_parent=False):
 
         if return_parent:
             return parent, desired
-        return desired
+        return parent[desired]
```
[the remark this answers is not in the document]
wetzelj: Sorry for the delay responding... this took a bit to research:

**Level 1 Fields (Fields not nested within Sequences)**
It won't return a key error because by the time we get to this point, we're ensured that the tag exists in the header because expand_field_expression either expanded the expander into a list of tags, expanded the values/fields expression, or in the case of a single field being specified by the recipe, returned a dictionary containing the single field.  We don't attempt to run the action if expand_field_expression returns an empty list - no header fields identified.

That said, get_nested_field is used in only two places:

1. delete_field
2. blank_field

In blank_field, we expect that the value returned from get_nested_field is a DataElement.  If we return desired, the value is a BaseTag, while delete utilizes the parent with the Tag index to delete the field.  If we want to keep the return type of "desired" consistent (which I actually prefer), see what I just committed. 

**Nested Fields (Sequences)**
I think we have a bigger problem here. When attempting to BLANK a field nested within a sequence (example: ```(0040, 0275)__0__(0008, 0050)```), the index between the two tags is causing us to return the tuple from expand_field_expression with parent = None.

For our efforts we generally just remove sequences, so handling this isn't a high priority, and honestly when I think it is handled we need to look at sequences as a whole.  Would you be okay with punting on this portion of it for now?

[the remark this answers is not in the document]
vsoch: Sure! If the problem exists in the current code, then this PR still makes the base better (and we can work on that next). Could you open an issue describing it in detail (likely copy pasting most of the above) so we don't forget?

And so I don't forget like last time - let's bump the version and add a note to the changelog! if you want to release this right away vs. wait for the fix for the above, it's up to you. Let me know!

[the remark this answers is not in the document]
wetzelj: Version and changelog are already bumped. :)  

I got a little sidetracked today.  Discovered that I was looking at an old version of the DICOM spec and was missing some VRs from my unit tests (hence today's update to the deid-data).  I'm pretty confident now that everything is covered.  Just committed some additional changes so that now all VRs are allowed for BLANK and covered by the unit tests as well.

vsoch:
Woot! passing now, and these tests are fantastic! @wetzelj when you are ready for merge, it shall be so.

wetzelj:
Thanks!  I'm working on a fix for #244 as well that I'd like to get included in this PR (along with some new tests for REPLACE which mirror what I did for BLANK).

vsoch:
@wetzelj you're amazing!!

deid/utils/actions.py
```diff
@@ -26,7 +27,7 @@ def parse_value(dicom, value, item=None, field=None, funcs=None):
         item = dict()
 
     # Does the user want a custom value?
-    if re.search("[:]", value):
+    if re.search("(^var:)|(^func:)|(^deid_func:)", value):
```
wetzelj, line 30: @vsoch - Modified this to be explicit.  By only searching for the colon, we were incorrectly sending URL/URI values down this path.  Noticed this with the addition of the tests for the UR VR.

vsoch: doh. Thanks for the catch.

wetzelj:
@vsoch - I think I'm set on what I'd like to get in this PR, but am interested in your feedback if there's anything that you'd like done differently.

.github/workflows/main.yaml
```diff
@@ -40,5 +40,5 @@ jobs:
         run: |
           export PATH="/usr/share/miniconda/bin:$PATH"
           source activate testing
-          pip install deid-data==0.0.17
+          pip install deid-data==0.0.19
```
vsoch, line 43: Do we need to pin this? I think we'd be better off always using the most up to date.

deid/utils/actions.py
```diff
@@ -66,10 +67,28 @@ def parse_value(dicom, value, item=None, field=None, funcs=None):
             # item is the lookup, value from the recipe, and field
             # The field is an entire dicom element object
             return item[value_option](dicom=dicom, value=value, field=field, item=item)
-
-        bot.warning("%s is not a valid value type, skipping." % (value_type))
-        return None
-    return value
+        else:
+            bot.warning("%s is not a valid value type, skipping." % (value_type))
```
vsoch, line 71: If we want to start using format strings across the code base, maybe it makes sense to only add string parsing using them.
```suggestion
            bot.warning(f"{value_type} is not a valid value type, skipping.")
```

deid/utils/actions.py
```diff
@@ -66,10 +67,28 @@ def parse_value(dicom, value, item=None, field=None, funcs=None):
...
+            return None
+
+    # Determine if the value is for an existing field.  If so,
+    # the value must be converted to conform to the appropriate Python type.
+    # Otherwise the field can remain as string and be auto-added as such.
+    if isinstance(field, str) and dicom is not None and field in dicom:
```
vsoch, line 77: The pattern I typically do here is have the default be false, and then the block set to true to avoid that extra "else" e.g.,

```python
existingField = False
if case1:
    existingField = True
elif case2:
    existingField = True
```

deid/utils/actions.py
```diff
@@ -66,10 +67,28 @@ def parse_value(dicom, value, item=None, field=None, funcs=None):
...
+        existingField = True
+        fieldName = field.name
+        fieldVR = field.element.VR
+    else:
+        existingField = False
+
+    if existingField:
+        return convert_value(fieldName, fieldVR, value)
+    else:
+        return value
```
vsoch, line 91: If you want a one-liner:
```python
return convert_value(fieldName, fieldVR, value) if existingField else value
```

deid/utils/actions.py
```diff
@@ -141,3 +160,41 @@ def get_timestamp(item_date, item_time=None, jitter_days=None, format=None):
...
+            return None
+    elif VR in ["SL", "SS", "SV", "UL", "US", "UV"]:
+        try:
+            return int(value)
+        except (ValueError, TypeError):
+            bot.warning(
+                f"Value ({value}) is not a valid value for VR {VR} field: {field}. Field will be BLANKED."
+            )
+            return None
+    else:
```
vsoch, line 199: I don't think we need the else - if we reach the end of the function, just return value, as all the checks above return something.

.github/workflows/main.yaml
wetzelj: Not from my perspective.  It was already pinned, so I just updated the version.  Committed the change.

deid/utils/actions.py
wetzelj: I'm wordy. :)  It's all updated.

wetzelj:
Adding #243 to this PR as well.  Please hold off merging. New commits will arrive tomorrow.

deid/dicom/parser.py
```diff
@@ -160,8 +160,9 @@ def get_nested_field(self, field, return_parent=False):
             # Otherwise it's an index into a sequence
             else:
```
wetzelj, line 162: Here's an explanation of the change below:

If we fall into the else, the uid is a sequence.  We don't want to check for it's presence in a dictionary, but instead want to check to ensure that the uid (index) is within the bounds of the array of items in the SQ DataElements value which is a Sequence - an iterable.  If the index is outside the bounds we return None as the parent.  If the index is within the bounds of the array, we want to return the DataElement at that index (the specific occurrence of the 
sequence element) for update.

deid/tests/test_replace_identifiers.py
```diff
@@ -118,16 +118,12 @@ def test_replace_with_constant(self):
         ]
         recipe = create_recipe(actions)
```
wetzelj, line 120: Explanation of the change below:

After implementing the change for sequence processing, this unit test was failing with an index out of bounds exception.  This exception was occurring because the DicomParser object was applying the stock deid.dicom recipe which was causing the Accession Number field to be removed.  I did not dig into why the sequence change affected this unittest in this manner, but since it was only impacting the check for getting the "before" values and all other unit tests passed, I didn't feel that it warranted further investigation. Instead I just updated this to red the file and directly retrieve the "before" values

wetzelj:
@vsoch - This wraps up what I'm intending (plus a little more) for this PR.  Once you're good with the inclusions, please go ahead with the merge at your convenience.

vsoch:
I just happen to be up since 6am debugging registry authentication for another OSS project - trying to squeeze in before the workday. Hashtag, maintainer life! :laughing:

Issue #253, opened 2023-05-02
How to specify Private tags in the recipes


howff:
I'm trying to specify a recipe for a specific type of file (from the Ziehm manufacturer) which doesn't fill in the `Manufacturer` field, only some Private tags. I'm sure that I don't fully understand deid, nor Private tags, but I can't get it to work.

```
LABEL Ziehm
  contains 0x00190012 ZIEHM_1.0 ImageCaptureData
  ctpcoordinates 0,0,196,148
```

Am I going about this the wrong way, should I specify Private tags differently?

I can see an easy fix, in `filter.py` function `apply_filter` simply
```
if '0x' in field: field = int(field, 0)
```

vsoch:
This could be we are missing that line! Pinging @wetzelj and if you try this out (and it fixed the issue) feel free to open a PR.

wetzelj:
Interesting. I agree that enabling targeting of private tags in filter sections is something that should be handled and available.  While it's not come up as a production use case for us, I could see it happening. 

Just glancing at the code, I'm wondering if converting the field to an int will be enough.  Does this correctly retrieve the private tag field in all situations?  This change would be a great opportunity to add some private tag unit tests to [test_filter_detect.py](https://github.com/pydicom/deid/blob/master/deid/tests/test_filter_detect.py).  These tests were initially added focusing on the interplay between two conditions within a rule (#142), but it would be beneficial to add more atomic tests for private/public tags and each of the filter types.

vsoch:
@howff it sounds like this would be a good addition - would you like to take a shot at the PR? @wetzelj has given direction to the file that we would want to add tests in, and the type of tests to add. Let us know if you have any questions!

howff:
The rationale for this is that some machines don't fill in the Manufacturer or Model tags so I have to check some Private tags to try and guess what they are. I'm only suggesting adding the ability to specify a tag by number instead of name, in the format `0xGGGGGEEEEE` so it could be used to refer to any tag not just private ones. Ideally one could find a private group by name (i.e. the value of a Private Creator) but that I believe is the subject of a different issue #205 

I'll see if I can make a PR.

howff:
Can this issue be closed now?

vsoch:
yes!

Issue #205, opened 2022-03-18
Finding relocated private elements using "PrivateCreator"


moloney:
Private elements can be relocated (group number will change) and some popular PACS will absolutely do this to data passing through them.  Instead of looking up an absolute `(group_number, element_number)` in rules, we should allow looking up private elements with a `PrivateCreator` plus an `element_number`. Then when processing a dataset you look for a matching `PrivateCreator` string (e.g. "SIEMENS CSA HEADER") to find the private group number dynamically.

I'm not sure how this capability should be exposed in the recipes, any thoughts?

vsoch:
Can you give me a little example in pseudo code? E.g., the idea is that the PrivateCreator is a string and it's use in place of the group number?

wetzelj:
@moloney - Would a %values section work for you? 

With a values section, you can define a field (or list of fields) that you want to use to define a set of values.  This values list would then be used in conjunction with a REMOVE or other command to change the identified tags.

```
%values manufacturer_values
FIELD Manufacturer

%header
REMOVE values:manufacturer_values
```

**Header Before**
PatientName - John Doe
Manufacturer - Siemens
(0009,0009) - SIEMENS CSA HEADER

**Header After**
PatientName - John Doe

If that doesn't work, a custom function may be another option for you.  The custom function could be written to be grab the value of the tag being acted on, look for "SIEMENS CSA HEADER" and then return the appropriate value for the action that you want to take.

jstorrs:
I don't know if this background information is helpful or not by private elements are tricky in DICOM because it's designed to allow multiple vendors to add private tags without stepping on each other's toes (if they do it correctly). Often people get things directly from modalities and don't encounter these things.

Generally we think of private tags as similar to DICOM standard tags as having fully-specified group element pairs. But for private tags there's a layer of indirection.

For example here is a Siemens CSA group (used a lot in MRI research):

```
(0029,0010) LO [SIEMENS CSA HEADER]                     #  18, 1 PrivateCreator
(0029,0011) LO [SIEMENS MEDCOM HEADER2]                 #  22, 1 PrivateCreator
(0029,1008) CS [IMAGE NUM 4]                            #  12, 1 Unknown Tag & Data
(0029,1009) LO [20211010]                               #   8, 1 Unknown Tag & Data
(0029,1010) OB 53\56\31\30\04\03\02\01\65\00\00\00\4d\00\00\00\45\63\68\6f\4c\69... # 11008, 1 Unknown Tag & Data
[...]

PR #254, opened 2023-05-03
filter.py - Allow tags to be specified using a string containing a 0x-prefix hex number

howff:
# Description

Allow tags to be specified using a string containing a 0x-prefix hex number

Related issues: #253 

Converts a field name which is a string in the form '0xGGGGEEEE' (GroupElement) into a number so that it can be found in the dataset dict. This allows recipe rules to specify tags numerically. One use for this is to filter on private tags.

# Checklist

# Open questions

Need to consider the implications of #205 (Finding relocated private elements using "PrivateCreator")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+- Allow filter tag names to be 0x-prefix hex numbers so private tags can be referenced in recipes [#253]
 - Fix incorrect coordinate definition for GE CT [#249](https://github.com/pydicom/deid/issues/249)
 - Circular import error [#247](https://github.com/pydicom/deid/issues/247) (0.3.21)
 - Expand BLANK Action to additional VRs [#241](https://github.com/pydicom/deid/issues/241) (0.3.2)
--- a/deid/dicom/filter.py
+++ b/deid/dicom/filter.py
@@ -24,11 +24,14 @@ def apply_filter(dicom, field, filter_name, value):
     Parameters
     ==========
     dicom: the pydicom.dataset Dataset (pydicom.read_file)
-    field: the name of the field to apply the filter to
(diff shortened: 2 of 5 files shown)


howff:
I've also updated the documentation, CHANGELOG, and added a new test (with associated recipe file).

vsoch (commented):
This is a great start! We will want to bump the version. Also ping @wetzelj to discuss the tests he thinks would be good to add.

vsoch:
This overall looks great - the formatting is failing because you likely don't have `black==23.3.0 ` locally. My one question is if we want a test or two more? From @wetzelj:

> Just glancing at the code, I'm wondering if converting the field to an int will be enough. Does this correctly retrieve the private tag field in all situations? This change would be a great opportunity to add some private tag unit tests to [test_filter_detect.py](https://github.com/pydicom/deid/blob/master/deid/tests/test_filter_detect.py). These tests were initially added focusing on the interplay between two conditions within a rule (https://github.com/pydicom/deid/issues/142), but it would be beneficial to add more atomic tests for private/public tags and each of the filter types.

Specifically this part:

> it would be beneficial to add more atomic tests for private/public tags and each of the filter types.

@wetzelj do you think more tests are warranted here, or was this scoped to another follow up PR?

howff:
I've added a test specifically for the new feature, so I guess if you want different tests then maybe they could be in a different issue/pr?

wetzelj:
I haven't had a chance to look at this today and don't expect to be able to circle back to it until Friday, but I would also contend that it's important to test that this new feature didn't unintentionally break existing expected functionality.

wetzelj:
This should be okay.  Given that the other tests check the string method, they're already validating existing functionality.  Thanks for your work @howff!

vsoch (approved):
Thank you @howff !

wetzelj:
@vsoch, @howff  - Can this be merged and completed?

vsoch:
This was my bad! I was probably waiting for tests to finish and my browser crashed and I forgot to check again. Thanks for the reminder @howff.

vsoch:
Thank you! https://pypi.org/project/deid/0.3.22/

PR #255, opened 2023-06-14
Update docstrings formatting refactor

bushjam1:
# Description

This PR submits several minor formatting and docstring changes and one change to the logic of a function using multiple conditionals to use a dictionary instead. No changes are part of a current issue.  

Related issues: # (issue)

None

There are changes to several docstrings in the file (tags.py) to add parameters, formatting to reduce line lengths >100 chars, and one change to the function apply_filter() in filter.py to use dict map instead of conditionals.  

# Checklist

- [X ] I have commented my code, particularly in hard-to-understand areas

# Open questions

None. Thanks for reviewing!

### DIFF
--- a/deid/dicom/filter.py
+++ b/deid/dicom/filter.py
@@ -26,37 +26,27 @@ def apply_filter(dicom, field, filter_name, value):
     dicom: the pydicom.dataset Dataset (pydicom.read_file)
     field: the name of the field to apply the filter to,
       or the tag number as a string '0xGGGGEEEE'
-    filer_name: the name of the filter to apply (e.g., contains)
+    filter_name: the name of the filter to apply (e.g., contains)
     value: the value to set, if filter_name is valid
 
     """
     if "0x" in field:
         field = int(field, 0)  # 0=decode hex with 0x prefix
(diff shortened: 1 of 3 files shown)


deid/dicom/filter.py
```diff
@@ -26,37 +26,27 @@ def apply_filter(dicom, field, filter_name, value):
...
       or the tag number as a string '0xGGGGEEEE'
-    filer_name: the name of the filter to apply (e.g., contains)
+    filter_name: the name of the filter to apply (e.g., contains)
     value: the value to set, if filter_name is valid
 
     """
     if "0x" in field:
         field = int(field, 0)  # 0=decode hex with 0x prefix
     filter_name = filter_name.lower().strip()
+    filter_name_map = {
```
vsoch, line 36: I see what you are trying to do, but this design isn't an improvement because in creating the dictionary you are executing each function. We definitely don't want to do that.

bushjam1: Thank you for the feedback.

vsoch: If you’d like to revert this change, the docstring fixes are great and would be much appreciated!

Issue #266, opened 2024-10-01
Handle breaking change with pydicom 3


sammaxwellxyz:
recent pydicom release includes breaking change

https://github.com/pydicom/pydicom/releases/tag/v3.0.0

the install requires needs an upper bound or the code updated, a particular breaking change is the attempted use of `read_file` from dicom/header.py

sammaxwellxyz:
Nice, will put that together, in the meantime I created a PR for capping the version if your prefer https://github.com/pydicom/deid/pull/267, by all means dismiss :)

trex-arth:
Can this issue be closed?

PR #267, opened 2024-10-01
fix: define max version for pydicom to avoid breaking changes

sammaxwellxyz:
# Description

Related issues: #266 https://github.com/pydicom/deid/issues/266

define install requires to include upper bound for pydicom.

the preference may be for an update to adopt pydicom 3 but this may be a useful fix in the short term as that may introduce breaking changes

# Checklist

- [x] My code follows the style guidelines of this project (i think)

# Open questions

you may want a similar upper bound for numpy, accidentally adopting 2+ has been an issue for some

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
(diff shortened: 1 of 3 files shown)


vsoch:
Good thinking to get this in for a closer version (before we bump up). Could you please bump the version in deid/version.py and put a note in the changelog?

vsoch:
@sammaxwellxyz looks like there is just a linting issue - we have it pinned to black 23.3.0, so either running that version or updating the pin (and running the version you choose on the codebase) would work here. Thanks!

Issue #275, opened 2025-03-15
`deid` CLI tool never saves dicom output


WillForan:
I've maybe misunderstood the intended workflow of died's `main`. I expected to run
```
deid --debug --outfolder /tmp/phi-clean/ --format dicom --overwrite identifiers --action all --deid ./deid-py.cfg --input phi/
```
and see new dicoms in `/tmp/phi-clean`. But the folder is left empty.

When `put` is requested for action `identifiers`, `replace_identifiers` is called but the default `save` is never changed, left as the default `False`. While the script ending output message is "files at {output_folder}", that folder will never get new dicoms(?)

https://github.com/pydicom/deid/blob/master/deid/main/identifiers.py#L66
``` 
    cleaned_files = replace_identifiers(
            dicom_files=dicom_files,
            ids=ids,
            deid=args.deid,
            overwrite=args.overwrite,
            output_folder=output_folder,
        )
```

https://github.com/pydicom/deid/blob/master/deid/dicom/header.py#L91
```
def replace_identifiers(
    dicom_files,
    ids=None,
    deid=None,
    save=False,
    ...
```

vsoch:
The `--overwrite` flag means that you will write over the input dicoms. Do you see changes there?

WillForan:
ohhh! hmmm

If I start fresh and don't use `--overwrite`, I still don't see any files in my `/tmp/phi-clean` 

But if I use a modified version (#276) with of `identifiers.main` where `replace_identifiers(save=True)` and run the same command twice without `--overwrite` I think I see an expected error:
>  ERROR phi.dcm already exists, overwrite set to False. Not writing.

-----

```
# clear out any files
rm /tmp/phi-clean/*

# modified replace_identifiers(save=True)
deid --debug --outfolder /tmp/phi-clean/ --format dicom identifiers --action all --deid ./deid-py.cfg --input phi/ 
GET and PUT identifiers from 
1 dicom files at /tmp/phi-clean/

# rerun with out overwrite, get expected error
deid --debug --outfolder /tmp/phi-clean/ --format dicom identifiers --action all --deid ./deid-py.cfg --input phi/ 
GET and PUT identifiers from 
ERROR phi.dcm already exists, overwrite set to False. Not writing.
1 dicom files at /tmp/phi-clean/
```

vsoch:
That looks right! See my comment in your PR.

vsoch:
Closed with #276

PR #276, opened 2025-03-15
fix: main.identifiers saves dicom #275

WillForan:
new test file to confirm cli usage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+- `deid` client saves cleaned dicoms when requested [#276](https://github.com/pydicom/deid/pull/276) (0.4.1)
 - Update to use pydicom 3 [#267](https://github.com/pydicom/deid/pull/267) (0.4.0)
 - Refactor INCLUDE_REQUIRES and provide max pydicom version [#267](https://github.com/pydicom/deid/pull/267) (0.3.25)
 - Support pydicom.Dataset objects created from BytesIO [#265](https://github.com/pydicom/deid/pull/265) (0.3.24)
--- a/deid/main/identifiers.py
+++ b/deid/main/identifiers.py
@@ -69,6 +69,7 @@ def main(args, parser):
             deid=args.deid,
             overwrite=args.overwrite,
             output_folder=output_folder,
+            save=True,
         )
 
         bot.info("%s %s files at %s" % (len(cleaned_files), args.format, output_folder))
--- a/deid/tests/test_cli.py
+++ b/deid/tests/test_cli.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+
+import os
+import shutil
+import tempfile
+import unittest
+from unittest.mock import patch
+
+import deid.main
+from deid.data import get_dataset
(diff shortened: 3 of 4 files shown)


vsoch:
I see what you are finding - it looks like save defaults to False (the normal usage of deid is within Python, where you don't overwrite / save files but get them back from the function call). We definitely would want save to be True for the client case, because you can't do anything with returned objects in a function that then exits!

For a sanity check, can we try a run as it is to confirm it fails?

WillForan:
yeah! Not sure what the best way to do that is. Here's another branch with the failing test
https://github.com/WillForan/deid/tree/cli-save-fail (I can PR on that branch and then put back the save=True?)

```
 python -m unittest deid/tests/test_cli.py

######################START######################
GET and PUT identifiers from 
WARNING No deid specification provided, will use defaults.
1 dicom files at out/
E
######################END########################

======================================================================
ERROR: test_deidmain_write_identifiers (deid.tests.test_cli.TestMainAction.test_deidmain_write_identifiers)
Run example command line call. Expect saved output.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/foranw/.local/share/uv/python/cpython-3.12.8-linux-x86_64-gnu/lib/python3.12/unittest/mock.py", line 1396, in patched
    return func(*newargs, **newkeywargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/foranw/src/work/utils/deid/deid/tests/test_cli.py", line 47, in test_deidmain_write_identifiers
    outfile = utils.dcmread("out/example.dicom")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/foranw/src/work/utils/deid/deid/dicom/utils.py", line 117, in dcmread
    return pydicom.dcmread(filename, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/foranw/src/work/utils/deid/.venv/lib/python3.12/site-packages/pydicom/filereader.py", line 1042, in dcmread
    fp = open(fp, "rb")
         ^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'out/example.dicom'

----------------------------------------------------------------------
Ran 1 test in 0.177s

FAILED (errors=1)

```



```
git diff cli-save
diff --git a/deid/main/identifiers.py b/deid/main/identifiers.py
index 10472dc..00e3fe8 100644
--- a/deid/main/identifiers.py
+++ b/deid/main/identifiers.py
@@ -69,7 +69,6 @@ def main(args, parser):
             deid=args.deid,
             overwrite=args.overwrite,
             output_folder=output_folder,
-            save=True,
         )
 
         bot.info("%s %s files at %s" % (len(cleaned_files), args.format, output_folder))


git switch cli-save
python -m unittest deid/tests/test_cli.py

######################START######################
GET and PUT identifiers from 
WARNING No deid specification provided, will use defaults.
1 dicom files at out/

######################END########################
.
----------------------------------------------------------------------
Ran 1 test in 0.183s

OK
```

vsoch:
I can see the output above and the logic in the code, so no need. I think this is good to go - if you could please bump the version in version.py and add a corresponding note in the CHANGELOG.md we should be good.

WillForan:
Sure! I don't have a good grip on semver. This will be `4.0.1`?

vsoch (approved):

vsoch:
Thank you @WillForan! 

https://pypi.org/project/deid/0.4.1/

Issue #284, opened 2025-09-03
CLI does not run `clean_pixel_data` / `DicomCleaner`?


WillForan:
running
```
deid --debug --overwrite -o /tmp/phi-clean identifiers -a all --deid ./deid-py.cfg --input phi.dcm 
```

with `deid-py.cfg` like
```
FORMAT dicom

%filter greylist

LABEL Censor Top Left
contains SOPInstanceUID .
  coordinates 0,0,512,110

%header
ADD PatientIdentityRemoved YES
```

We can see the header change, but no voxels are modified, `clean_pixel_data` doesn't run (?)

```bash
diff <(pydicom show  phi.dcm) <(pydicom show phi-clean/phi.dcm )
# 32a33
# > (0012, 0062) Patient Identity Removed            CS: 'YES'
```

```python
from pydicom import read_file as dcmread; import numpy as np
np.all(dcmread('phi.dcm').pixel_array == dcmread('phi-clean/phi.dcm').pixel_array)
# np.True_
```

from [dicom-pixels.md](https://pydicom.github.io/deid/getting-started/dicom-pixels/) and [run-cleaner-client.py](https://github.com/pydicom/deid/blob/master/examples/dicom/pixels/run-cleaner-client.py), I gather `DicomCleaner` would need to be invoked in addition to or instead of `replace_identifiers` in [`deid/main/identifiers.py`](https://github.com/pydicom/deid/blob/master/deid/main/identifiers.py#L66) ?

WillForan:
Oh! Am I just missing a command line argument?

If there's not one yet and we want to use the cli, is the next step to add something like a `args.command == "pixels"` to [`main/__init__.py`](https://github.com/pydicom/deid/blob/master/deid/main/__init__.py#L184)?

I imagine cleaning pixels and headers at the same time is useful. Doing that seems tricky: need to pass modified files from `DicomParser` to `DicomCleaner (?)
That might look having `replace_identifiers`  save into a temporary directory and to be the input for DicomCleaner. Am I making this too hard -- is there a simple way to clean pixels and header identifiers?

vsoch:
I don't think we ever exposed a command line client to do the cleaning, just replace identifiers. I'd be happy to accept a PR to add one if you think it makes sense.

PR #285, opened 2025-09-04
CLI 'pixel' command

WillForan:
Minimal implementation of a `clean-pixels` command option for the CLI (#284).

  * [deid/dicom/pixels/clean.py:_get_clean_name](https://github.com/pydicom/deid/pull/285/files#diff-01559f95064ff2dac0358b74c933704bd1dd024b1a74b4e8eb84ebc41a1adade): refactor to allow no `clean-` prefix and no extension change for predicable dicom_save (new directory only, not new names). Not sure if that's wise -- just how I'd originally written the test and another user might expect output data.
  *  [deid/main/pixels.py](https://github.com/pydicom/deid/pull/285/files#diff-0f86b2e0af5064fa29b34350dee26bfa7fb6149eb25f24924fd9646357b616f6): new file runs `DicomCleaner` detect, clean, save_dicom on input directory

I think it'd be useful to do both header and pixel scrubbing in one interface so everything described in a deid profile/config file can be applied at once. I'm not sure yet what the best way to go about that is (run identifiers into a new temp dir, rerun clean on that and save output to final location?)

### DIFF
--- a/deid/main/__init__.py
+++ b/deid/main/__init__.py
@@ -18,7 +18,7 @@ def get_parser():
         description="Deid (de-identification, anonymization) command line tool."
     )
 
-    # Global Variables
+    # Global Variables: generic options available for all 'actions' command
     parser.add_argument(
         "--quiet",
         "-q",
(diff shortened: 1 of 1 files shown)


deid/dicom/pixels/clean.py
```diff
@@ -233,17 +244,35 @@ def animate(i):
         else:
             bot.warning("use detect() --> clean() before saving is possible.")
 
-    def save_dicom(self, output_folder=None, image_type="cleaned"):
+    def save_dicom(
```
vsoch, line 247: The save function should be agnostic to the file it is being given (clean or not) - so the `image_type` doesn't make sense here. Any logic for naming the file should be in the calling function.

deid/dicom/pixels/clean.py
```diff
@@ -153,8 +157,15 @@ def _get_clean_name(self, output_folder, extension="dcm"):
             bot.debug("Creating output folder %s" % output_folder)
             os.makedirs(output_folder)
 
-        basename = re.sub("[.]dicom|[.]dcm", "", os.path.basename(self.dicom_file))
-        return "%s/cleaned-%s.%s" % (output_folder, basename, extension)
+        # do we want to change extension? this might convert .dicom to .dcm
```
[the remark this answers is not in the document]
vsoch: Ha! That's interesting indeed.

My sense is that renaming of files is outside of the jurisdiction of a cleaning tool. I think it's reasonable to add a clean prefix, but (unless it is a convention to not have an extension) I don't think deid should be responsible for the convention of a specific center.

deid/dicom/pixels/clean.py
vsoch: Let's try for that.

WillForan: At first it looks like changing this could be a breaking change. Someone downstream could be using `DicomClient().save_dicom(image_type="original")` (or more likely `save_png(image_type="original")`. But best I can tell, `self.original` is at least currently not an attribute. So only thing this breaks is non-working code? ... maybe :shrug:

WillForan:
(hopefully) fixed the redundant `--deid` and `--input`. confirmed command works as expect on a dataset in the wild as well as in `test_cli.py` with example deid-data

Incidentally, `test_cli.py` also caught a failed attempt to move `--deid` into the global `parser` instead of every subparser. Order is important and moving to global would break the existing specification (currently have eg. `deid identifiers --deid deid.cfg` but into global would force `--deid` before the sub command like `deid --deid deid.cfg identifiers`)

PR #290, opened 2025-09-23
Allow saving with a compressed transfer syntax

ReeceStevens:
# Description

Related issues: None

Currently, `deid` only saves images in decompressed format. This can lead to some surprising balloons in disk usage when you expand compressed DICOM images. In my particular use case, I noticed this when running `deid` in Pyodide due to a sharp increase in RAM usage-- this correlated to saving DICOM in decompressed format to the memFS filesystem, forcing the WASM memory allocation to increase substantially.

This PR introduces two new options to the `save_dicom` function-- `preserve_compression`, which will attempt to save the dataset back into the same compression format it had originally; and `compression`, which allows the user to specify a particular compression transfer syntax they would like to use for the output.

I understand that adding additional kwargs is not always desirable-- I think the `compression` kwarg is the most critical for my own use case, so if you would prefer to toss `preserve_compression` I can revert that one.

# Checklist

# Open questions

Questions that require more discussion or to be addressed in future development: None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
+- Allow saving with a compressed transfer syntax [#290](https://github.com/pydicom/deid/pull/290) (0.4.7)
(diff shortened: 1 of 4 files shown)


ReeceStevens:
Thanks for the review! Changelog and version bump have been applied. I also added a unit test exercising the saving of files with a user-specified compression transfer syntax.

vsoch (approved):
This is excellent! I'll merge this in (and not release) and we can finish up your other PR for the same release. I may not get to it tonight, but soon!

PR #293, opened 2025-11-17
bug: Fix field removal and blanking to clean up child UID references

Simlomb:
# bug: Fix field removal and blanking to clean up child UID references

In the DICOM anonymization workflow, the parser maintains an internal lookup ([self.fields](vscode-file://vscode-app/snap/code/211/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) that maps unique identifiers (UIDs) to DICOM fields. These UIDs represent both top-level tags and nested tags (such as those inside sequences).

## Before the fix:

- When you remove or blank a parent sequence tag (for example, a tag representing a DICOM sequence), the actual DICOM data structure deletes or blanks the parent tag and all its child elements.
- However, the internal lookup ([self.fields](vscode-file://vscode-app/snap/code/211/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) still retained entries for the child tags that were previously nested under the parent sequence.
- This means that, although the DICOM data no longer contained those child tags, the parser’s internal cache still referenced them.
- If a subsequent recipe action (such as REPLACE, BLANK, REMOVE, etc.) tried to operate on one of these child tags, the parser would find the UID in its internal lookup and attempt to perform the action.
- When you remove or blank a parent sequence tag (for example, a tag representing a DICOM sequence), the actual DICOM data structure deletes or blanks the parent tag and all its child elements.
- Since the actual DICOM data no longer contained the child tag, this would result in error.

## After the fix:

- When you remove or blank a parent sequence tag, all child UIDs nested under that parent are also removed from self.fields.
- Subsequent actions do not find stale child references, preventing errors and ensuring consistency.

## Summary of PR: 

- Ensure child field UIDs are removed from internal lookup when parent sequence is deleted or blanked
- Update tests in test_action_interaction.py to show this behavior for blanked and removed sequences
- Prevent stale references and errors in subsequent recipe actions

# Checklist

### DIFF
--- a/deid/dicom/parser.py
+++ b/deid/dicom/parser.py
@@ -176,6 +176,17 @@ class DicomParser:
             return parent, desired
         return desired
 
+    def get_child_fields(self, field):
+        """
+        Return a list of child field UIDs for a given field.
+
+        This method identifies all field UIDs in self.fields that are nested under the provided field,
+        based on the UID prefix convention (parent UID + '__').
+        It is used to find and remove all child fields when blanking or deleting a parent field (e.g., a sequence).
+        """
+        field_uid_prefix = field.uid + "__"
+        return [uid for uid in self.fields if uid.startswith(field_uid_prefix)]
+
     def delete_field(self, field):
         """
         Delete a field from the dicom.
@@ -191,11 +202,7 @@ class DicomParser:
             # Remove the field itself from the lookup
             self.fields.remove(field.uid)
             # Also remove any child fields that were nested under this field
-            field_uid_prefix = field.uid + "__"
-            child_uids = [
-                uid for uid in self.fields if uid.startswith(field_uid_prefix)
-            ]
-            for child_uid in child_uids:
+            for child_uid in self.get_child_fields(field):
                 self.fields.remove(child_uid)
 
     def blank_field(self, field):
@@ -211,11 +218,7 @@ class DicomParser:
         if parent and desired in parent:
             parent[desired].value = None
             # Also remove any child fields that were nested under this field
-            field_uid_prefix = field.uid + "__"
-            child_uids = [
-                uid for uid in self.fields if uid.startswith(field_uid_prefix)
-            ]
-            for child_uid in child_uids:
+            for child_uid in self.get_child_fields(field):
                 self.fields.remove(child_uid)
 
     def replace_field(self, field, value):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and **Merged pull requests**. Critical items to know are:
 Referenced versions in headers are tagged on Github, in parentheses are for pypi.
 
 ## [vxx](https://github.com/pydicom/deid/tree/master) (master)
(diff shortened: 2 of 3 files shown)


Simlomb:
@vsoch while using the tool on some data, I noticed that when one remove or blank a parent sequence tag (for example, a tag representing a DICOM sequence), the actual DICOM data structure deletes or blanks the parent tag and all its child elements, but the internal lookup (self.fields) still retained entries for the child tags that were previously nested under the parent sequence. This caused errors in the deidentification process if the child tag was also included in the recipe with another action. I fixed it in the fix proposed in this PR.
Please let me know if you have any comment on it.

deid/dicom/parser.py
```diff
@@ -181,22 +181,42 @@ def delete_field(self, field):
...
         # Returns the parent, and a DataElement (indexes into parent by tag)
         parent, desired = self.get_nested_field(field, return_parent=True)
         if parent and desired in parent:
             parent[desired].value = None
+            # Also remove any child fields that were nested under this field
+            field_uid_prefix = field.uid + "__"
+            child_uids = [
+                uid for uid in self.fields if uid.startswith(field_uid_prefix)
+            ]
+            for child_uid in child_uids:
```
vsoch: A few comments. When we blank a field, it technically is still there (just blank) so why would we remove it (and child fields) here? Also, if there is shared logic between these functions to derive and yield child fields, we might instead have a shared function:

```python
for child_uid in self.yield_child_fields(field):
    # do action(s)
```

vsoch (commented):
This looks good - with one question.

Apologies for delay - I was traveling since last Saturday and got back last night!

deid/dicom/parser.py
Simlomb: > A few comments. When we blank a field, it technically is still there (just blank) so why would we remove it (and child fields) here? 

When a parent field is blanked that tag has zero length, hence there are no more child tags. The error raised because the uid lookup still listed those nested tags with other actions, even when they did not exist anymore in the dicom file.  In those new lines we do not remove the tags from the dicom files, but just from the field lookup.

> Also, if there is shared logic between these functions to derive and yield child fields, we might instead have a shared function:
> 
> ```python
> for child_uid in self.yield_child_fields(field):
>     # do action(s)
> ```

Thanks for catching that, I added the new method. I also bumped the version and changelog. Let me know if you have any other comments.

deid/dicom/parser.py
```diff
@@ -176,27 +176,51 @@ def get_nested_field(self, field, return_parent=False):
...
+    def yield_child_fields(self, field):
+        """
+        Return a list of child field UIDs for a given field.
+
+        This method identifies all field UIDs in self.fields that are nested under the provided field,
+        based on the UID prefix convention (parent UID + '__').
+        It is used to find and remove all child fields when blanking or deleting a parent field (e.g., a sequence).
+        """
+        field_uid_prefix = field.uid + "__"
+        child_uids = [uid for uid in self.fields if uid.startswith(field_uid_prefix)]
```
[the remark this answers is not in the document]
Simlomb: Thank you for the suggestion! I’ve updated the method to use yield as you recommended. I completely agree, this approach is more appropriate than returning a full list.

PR #300, opened 2026-02-07
Added personal fingerprint to DicomField class

manalabumehsen:
Added a "modified_by" attribute in DicomField to include personal fingerprint for tracking modifications. Author: Manal Abu Mehsen


vsoch:
Thank you for this contribution, but we won't be adding it as a feature.

PR #302, opened 2026-02-19
feat: Add support for progress reporting

Simlomb:
feat: Add support for progress reporting

This PR introduces a callback mechanism to allow users to inject custom logic during processing. The callback enables real-time progress reporting, logging, or dynamic intervention without modifying the core workflow.

**Utility**:

- Decouples progress reporting or side effects from the main logic.
- Enables flexible integration with GUIs, CLI progress bars, or logging systems.
- Makes the codebase more extensible and testable.

# Checklist

### DIFF
--- a/deid/dicom/header.py
+++ b/deid/dicom/header.py
@@ -98,6 +98,8 @@ def replace_identifiers(
     strip_sequences=False,
     remove_private=False,
     disable_skip=False,
+    progress_callback_init=None,
+    progress_callback_update=None,
 ):
     """
     Replace identifiers.
@@ -118,7 +120,16 @@ def replace_identifiers(
 
     # Parse through dicom files, update headers, and save
     updated_files = []
+    # If a progress callback initializer is provided, call it with the total number of files.
+    # This can be used to set up a progress bar, e.g., tqdm(total=len(dicom_files)).
+    if progress_callback_init is not None:
+        progress_callback_init(len(dicom_files))
+
+    # Iterate through each DICOM file, calling the update callback if provided.
+    # The update callback can be used to update progress, such as incrementing a progress bar.
     for dicom_file in dicom_files:
+        if progress_callback_update is not None:
+            progress_callback_update(dicom_file)
         parser = DicomParser(
             dicom_file,
             force=force,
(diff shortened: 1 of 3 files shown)


Simlomb:
@vsoch this PR is simply to allow a way to obtain a progress status during the deidentification process. Please let us know if you have comments. 
I also bumped the version and CHANGELOG.

deid/dicom/header.py
```diff
@@ -118,7 +120,16 @@ def replace_identifiers(
...
     updated_files = []
+    # If a progress callback initializer is provided, call it with the total number of files.
+    # This can be used to set up a progress bar, e.g., tqdm(total=len(dicom_files)).
+    if progress_callback_init is not None:
+        progress_callback_init(len(dicom_files))
+
+    # Iterate through each DICOM file, calling the update callback if provided.
+    # The update callback can be used to update progress, such as incrementing a progress bar.
     for dicom_file in dicom_files:
+        if progress_callback_update is not None:
```
vsoch, line 131: Perhaps a better strategy would be to yield the updated file (instead of append to the list) and then the calling function can wrap this with tqdm? It seems messy to hard code this "run this callback" into the function. Arguably, as long as we yield a unit (and the user knows the number of total files) the progress bar should be easy to implement.

Simlomb: Thank you for your suggestion! I agree that yielding each updated file would make it easier to integrate with progress bars like tqdm. However, changing the behavior of replace_identifiers to yield results would break backward compatibility for users who expect a list.
What I propose: 
Instead of having all the logic inside replace_identifiers, I can introduce a new helper function, replace_identifiers_file, which handles processing a single DICOM file (including parsing, updating, and saving).
The main replace_identifiers function would simply iterate over the list of files and call replace_identifiers_file for each one, collecting the results in a list as before. This keeps the interface backward compatible.

With this structure, users who want more granular control (for example, to add a progress bar or custom logic) can now call replace_identifiers_file directly in their own loop. 

Would you agree with this solution?

vsoch: Why not just pass a single item list to the existing function then?

Simlomb: Indeed, for some reason the easiest solutions do not always come to mind first :D .
I would have avoided the if statements, addiitonal for loop and append operations, but they do not cause the code to considerably slow down even on milion of files. 
So I will take this approach and come back to the PR  if needed.

Simlomb: Everything seems ok with this approach so I close the PR. Thanks for the discussions!

